PR #30, opened 2022-04-04
feat: Adds octuple list, menu, dropdown components

yrchhabra-eightfold:


src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const Dropdowns = () => (
+    <>
+        <h2>Dropdown</h2>
+        <div style={{ display: 'flex', justifyContent: 'space-between' }}>
+            <Dropdown overlay={Overlay()}>
+                <DefaultButton />
+            </Dropdown>
+            <Dropdown overlay={Overlay()} trigger="hover">
+                <DefaultButton />
```
dylankilgore, line 20: Need to use `split={true}` and perhaps `splitButtonMenuProps` and `splitButtonAriaLabel`. You might consider how you created the pill component, but with Split buttons both sides are clickable. See: https://developer.microsoft.com/en-us/fluentui#/controls/web/button and note how split button behaves.

The dropdown is good to go. but would be awesome to get split button in too.

PR #39, opened 2022-04-05
chore: npm publish action

yrchhabra-eightfold:


.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,30 @@
+name: Publish Octuple
+
+on:
+  release:
```
dylankilgore, line 4: As discussed offline, update to tag push

dylankilgore (changes requested):

PR #55, opened 2022-04-12
feat: adds octuple panel component, add portal component

yrchhabra-eightfold:


dylankilgore (approved):

src/components/Panel/Panel.stories.tsx
```diff
@@ -0,0 +1,218 @@
+import React, { useState } from 'react';
+import { Panel, SizeType } from './';
```
dylankilgore, line 2: Don't forget to export

PR #58, opened 2022-04-13
feat: adds octuples dialog component

yrchhabra-eightfold:


src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,95 @@
...
+        height,
+        width,
+    };
+
+    useEffect(() => {
+        onVisibleChange?.(visible);
+    }, [visible]);
+
+    const getDialog = (): JSX.Element => (
+        <div
```
dylankilgore, line 63: `aria-modal="true"`

src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,95 @@
...
+            role="dialog"
+            className={dialogBackdropClasses}
```
dylankilgore, line 65: `aira-labelledby` using `id` of dialog `label` as its value - announces context of dialog in screen readers.

src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,95 @@
...
+                maskClosable && onClose(e);
+            }}
+        >
+            <div
+                className={dialogClasses}
+                style={dialogStyle}
+                onClick={stopPropagation}
+            >
+                <div className={headerClasses}>
+                    {header}
```
dylankilgore, line 76: The `header` area should contain a `label` element that's used by `aria-labelledby` above

dylankilgore (changes requested):

dylankilgore (approved):

PR #65, opened 2022-04-18
feat: adds octuples modal component

ychhabra-eightfold:


src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,44 @@
...
+    modalClassName,
+    modalWrapperClassName,
+    ...rest
+}) => {
+    const modalClasses: string = classNames([
+        styles.modal,
+        modalClassName,
+        { [styles.small]: size === ModalSize.small },
+        { [styles.medium]: size === ModalSize.medium },
+        { [styles.large]: size === ModalSize.large },
```
dylankilgore, line 22: Need Fullscreen modal too, there are many scenarios and existing features in the product that require this.

src/components/Modal/Modal.types.ts
```diff
@@ -0,0 +1,27 @@
+import { BaseDialogProps } from '../Dialog/BaseDialog/BaseDialog.types';
+
+export enum ModalSize {
+    small = 'small',
+    medium = 'medium',
+    large = 'large',
```
dylankilgore, line 6: fullscreen

dylankilgore (changes requested):

src/components/Dialog/BaseDialog/BaseDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+            <div
+                className={dialogClasses}
+                style={dialogStyle}
+                onClick={stopPropagation}
+            >
+                <div className={headerClasses}>
+                    <span id={labelId}>{header}</span>
+                    <NeutralButton icon={IconName.mdiClose} onClick={onClose} />
+                </div>
+                <div className={bodyClassName}>{body}</div>
```
dylankilgore, line 73: To make this scrollable you may consider adding a bodyMaxHeight prop. then refer to line 57 in the Modal.js common component in vscode `www/react/src/lib/common/Modal.js`. We either take the specified height or calculate it using the viewport, taking into account any variable insets and heights of the header and footer.

dylankilgore: You may also look at line 136 of the Modal common component scss to see the default calc for some ideas --> `www/react/src/styles/Modal.scss`

dylankilgore (changes requested):

dkilgore-eightfold (approved):

PR #69, opened 2022-04-19
chore: setup engines for node and npm version

yrchhabra-eightfold:


thiyagaraj-eightfold (approved):
Also lets a node badge in readme?

PR #72, opened 2022-04-19
chores: add chromatic

thiyagaraj-eightfold:
Add chromatic to the project


dkilgore-eightfold (approved):

ychhabra-eightfold (changes requested):

.github/workflows/build.yml
```diff
@@ -44,3 +44,13 @@ jobs:
           key: node_modules-${{ hashFiles('yarn.lock') }}
       - run: yarn
       - run: yarn lint
+  chromatic-deployment:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install dependencies
```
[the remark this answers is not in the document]
ychhabra-eightfold: yarn install also needs the correct node version

.github/workflows/build.yml
```diff
@@ -44,3 +44,15 @@ jobs:
           key: node_modules-${{ hashFiles('yarn.lock') }}
       - run: yarn
       - run: yarn lint
+  chromatic-deployment:
```
ychhabra-eightfold, line 47: We shouldn't be running this on the pull_request event but only on push event as recommend by the docs see https://www.chromatic.com/docs/github-actions#recommended-configuration-for-build-events

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,26 @@
+name: 'Chromatic'
+
+on: push
+
+env:
+  NODE_VERSION: 16
+jobs:
+  chromatic-deployment:
+    runs-on: ubuntu-latest
```
ychhabra-eightfold, line 9: No need to create another worklow yaml just add the following in the previous change you had done
```suggestion
    runs-on: ubuntu-latest
    if: github.event_name == 'push'
```

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+            fetch-depth: 0
+      - uses: actions/setup-node@v3
+      - name: Load cached node_modules
+        uses: actions/cache@v3
+        run: yarn
+        with:
```
ychhabra-eightfold, line 18: actions cache doesnt need node_version

.github/workflows/chromatic.yml
thiyagaraj-eightfold: Want to keep it separate, else it just becomes unmanageable yaml

.github/workflows/chromatic.yml
thiyagaraj-eightfold: its for yarn you mentioned

ychhabra-eightfold: i meant yarn needs 
- uses: actions/setup-node@v3
   with:
       node-version: ${{ env.NODE_VERSION }}

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,27 @@
...
+          path: node_modules
+          key: node_modules-${{ hashFiles('yarn.lock') }}
+      - name: Publish to Chromatic
+        uses: chromaui/action@v1
+        with:
+          node-version: ${{ env.NODE_VERSION }}
```
ychhabra-eightfold, line 26: dont think this is needed

ychhabra-eightfold:
the workflow was not correct, you shouldnt merge without approvals @thiyagaraj-eightfold

PR #75, opened 2022-04-19
feat: some faulty styling [don't merge]

thiyagaraj-eightfold:
just for test

### DIFF
--- a/src/components/Button/button.module.scss
+++ b/src/components/Button/button.module.scss
@@ -27,7 +27,7 @@
     --button-hover-variant-background-color: var(--primary-color-20);
     --text-color: var(--text-inverse-color);
     background-color: inherit;
-    border: 2px solid rgba(var(--button-primary-default-border-color), 0.01);
+    border: 8px solid rgba(var(--button-primary-default-border-color), 0.01);
     border-radius: $corner-radius-s;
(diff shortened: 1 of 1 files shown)


src/components/Button/button.module.scss
```diff
@@ -27,7 +27,7 @@
     --button-hover-variant-background-color: var(--primary-color-20);
     --text-color: var(--text-inverse-color);
     background-color: inherit;
-    border: 2px solid rgba(var(--button-primary-default-border-color), 0.01);
+    border: 8px solid rgba(var(--button-primary-default-border-color), 0.01);
```
dkilgore-eightfold, line 30: An 8px border is too thick.

dkilgore-eightfold (changes requested):

src/components/Button/button.module.scss
dkilgore-eightfold: This will cause a noticeable flicker on hover and will be jarring.

thiyagaraj-eightfold: This PR is just to test chromatic 😅

dkilgore-eightfold: Ah, oops.

PR #80, opened 2022-04-26
feat: button: adds two state button component

dkilgore-eightfold:


src/components/Button/TwoStateButton/TwoStateButton.tsx
```diff
@@ -0,0 +1,178 @@
...
+            buttonTextClassNames: string,
+            text: string
+        ): JSX.Element => (
+            <span className={buttonTextClassNames}>
+                {text ? text : 'Button'}
+            </span>
+        );
+
+        return (
+            <button
```
ychhabra-eightfold, line 150: Why arent we using BaseButton here ?

src/components/Button/button.module.scss
```diff
@@ -398,6 +448,66 @@
     }
 }
 
+.two-state-button {
```
ychhabra-eightfold, line 451: Can we move out two state button into its own scss file, its getting crowded here

src/components/Button/Button.types.ts
```diff
@@ -44,15 +55,36 @@ export interface SplitButtonProps
...
+     * The button counter number.
      */
-    type?: ButtonType;
+    counter?: number;
     /**
-     * Ref of the button
+     * The button icon 1.
      */
-    ref?: Ref<HTMLButtonElement>;
+    iconOne?: IconName;
```
ychhabra-eightfold, line 75: We also have Icon only two state buttons, for eg starring candidate. Will this component handle that, or that will be its own thing ?

ychhabra-eightfold (changes requested):

src/components/Button/button.module.scss
dkilgore-eightfold: It uses button classes, it's a button. Keeping it here should be necessary.

src/components/Button/TwoStateButton/TwoStateButton.tsx
dkilgore-eightfold: BaseButton is already overloaded, better to keep these separate.

ychhabra-eightfold (approved):

PR #82, opened 2022-04-27
fix: panel: updates panel close icon button to use neutral button

dkilgore-eightfold:


ychhabra-eightfold:
@dkilgore-eightfold Closing this pr i've fixed this in #83 as @ykelkar-eightfold pointed out another issue in panel to me yesterday.

PR #87, opened 2022-04-29
feat: configprovider: updates the config provider to accept custom styles

dkilgore-eightfold:
- moves the theme class to html tag
- adds use theme hook and its unit test
- updates the config provider story

Current draft only demonstrates the priority order of global and component styles from the ConfigProvider

Will need to add an example of a component prop theme.


src/components/ConfigProvider/Theming/Theming.types.ts
```diff
@@ -15,23 +15,198 @@ export type OcThemeNames =
...
+
+export interface OcInfoBarStyles extends OcButtonStyles {}
+
+export interface OcInputStyles extends OcButtonStyles {}
+
+export interface OcListStyles extends OcButtonStyles {}
+
+export interface OcMenuStyles extends OcButtonStyles {}
+
+export interface OcModalStyles extends OcDialogStyles {}
```
dkilgore-eightfold, line 111: we can refine where components should derive styles, for example Modal should extend Dialog, et cetera.

src/components/ConfigProvider/Theming/styleGenerator.ts
```diff
@@ -112,6 +112,14 @@ export function getStyle(themeOptions: ThemeOptions): IGetStyle {
         variables[`info-color`] = theme.infoColor;
     }
 
+    // ================= Border Radius ==================
+    if (theme.styles?.global?.borderRadius) {
```
dkilgore-eightfold, line 116: we might consider using a loop or recursive fn to make this more efficient.

PR #93, opened 2022-05-02
feat: change menu item value to type any

ychhabra-eightfold:


src/components/Menu/Menu.types.ts
```diff
@@ -1,7 +1,13 @@
 import { ButtonProps } from '../Button';
 import { ListProps } from '../List';
 
-export interface MenuItem extends Omit<ButtonProps, 'disruptive' | 'onClick'> {}
+export interface MenuItem
+    extends Omit<ButtonProps, 'disruptive' | 'onClick' | 'value'> {
+    /**
+     * Value of the menu item
+     */
+    value: any;
```
dkilgore-eightfold, line 9: Is there a better way to handle this? If this is to be a checkbox, or radio button - those components should be used? If it's an image + text, we might consider adding image source support to button like the input boxes.

dkilgore-eightfold (approved):

PR #97, opened 2022-05-02
feat: props: normalize label, classnames, button alignment and iconprops

dkilgore-eightfold:
This PR contains a BREAKING CHANGE:

classnames and icon props must be updated to match latest

Additionally:

- Moves Label into its own component
- Exposes Icon props like, spin and rotate
- Adds button alignIcon prop to render its icon on the left or right side of its text

### DIFF
--- a/src/__snapshots__/storybook.test.js.snap
+++ b/src/__snapshots__/storybook.test.js.snap
@@ -49,7 +49,7 @@ Array [
   <button
     aria-disabled={false}
     aria-label="Default Button"
-    className="button buttonDefault buttonPadding1"
+    className="button buttonDefault buttonPadding1 iconLeft"
     defaultChecked={false}
     disabled={false}
     onClick={[Function]}
@@ -88,7 +88,7 @@ Array [
   <button
     aria-disabled={false}
     aria-label="Default Button"
-    className="button buttonDefault buttonPadding1"
+    className="button buttonDefault buttonPadding1 iconLeft"
     defaultChecked={false}
     disabled={false}
     onClick={[Function]}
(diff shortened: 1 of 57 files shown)


thiyagaraj-eightfold:
reopening to trigger tests

ychhabra-eightfold:
@dkilgore-eightfold build is breaking cause of 

/home/runner/work/octuple/octuple/src/components/MatchScore/MatchScore.tsx(3,10)\n' +
[205](https://github.com/EightfoldAI/octuple/runs/6265916429?check_suite_focus=true#step:5:205)
    `      TS2305: Module '"../../shared/utilities"' has no exported member 'classNames'.\x1B[39m\x1B[22m\n` +
[206](https://github.com/EightfoldAI/octuple/runs/6265916429?check_suite_focus=true#step:5:206)
    ' @ ./src/components/MatchScore/index.ts 17:13-36\n' +

thiyagaraj-eightfold:
Hey - why don’t you try a component approach for this. 
example - `<Atom classNames={[...]} element='div' /> `

So `<Atom/>` or `<BaseElement/>` becomes a primitive component that can be used across the lib.
This will abstract how we are merging the classes etc.
We can always refactor the atom class in the future without really disturbing the API of the component

ychhabra-eightfold (approved):
LGTM

dkilgore-eightfold:
> Hey - why don’t you try a component approach for this. example - `<Atom classNames={[...]} element='div' /> `
> 
> So `<Atom/>` or `<BaseElement/>` becomes a primitive component that can be used across the lib. This will abstract how we are merging the classes etc. We can always refactor the atom class in the future without really disturbing the API of the component

Let's discuss this approach. Sounds like a good one. Maybe we could create a wrapping component.

PR #99, opened 2022-05-03
chore: add editconfig, vscode extension recommendations

phrough:


.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
```
ychhabra-eightfold, line 1: We use prettier, please setup prettier on ur editor and this wont be needed, will add this to contribution guidlines to make it clearer for future devs.

.vscode/extensions.json
```diff
@@ -0,0 +1,13 @@
+{
```
ychhabra-eightfold, line 1: I dont think any editor config/extenstions settings should part of this repo

ychhabra-eightfold (changes requested):

thiyagaraj-eightfold:
Hey @phrough , thanks for the initiative!
Can you create a github issue before creating a PR so that the team can comment and have discussions.

.editorconfig
phrough: Sounds good. Loosely related, there *might* be an issue in which prettier and eslint are battling each other work during the pre commit hook. I've seen some strange behavior there, but I'm not certain about it.

PR #100, opened 2022-05-03
fix: tabs: update tabs based on change in activeTab prop

ychhabra-eightfold:


src/components/Tabs/Tabs.context.tsx
```diff
@@ -12,6 +12,10 @@ const TabsProvider = ({ children, onChange, activeTab }: TabsContextProps) => {
     const [currentActiveTab, setCurrentActiveTab] =
         useState<TabValue>(activeTab);
 
+    useEffect(() => {
+        setCurrentActiveTab(activeTab);
+    }, [activeTab]);
+
```
thiyagaraj-eightfold, line 18: This is a bit confusing. If the state is managed locally then it should not be `activeTab`, rather it should be `initialActiveTab`.

thiyagaraj-eightfold: There can be two versions of the tabs component - 1. prop controlled, 2. with local state
This can be refactored a bit -
1. A props driven Tab component which takes `activeTab` 
2. Provide a useTabs hook which can be initialed with a default and manages all state

thiyagaraj-eightfold (changes requested):

src/components/Tabs/Tabs.context.tsx
ychhabra-eightfold: `activeTab` is passed as a prop and the used as the default activeTab and used to set the current active tab on mount. But this activeTab can change in different scenarios for eg: if a user goes back on a page, then the currentActiveTab is not updated in this situation. So for this reason if the activeTab changes we need to update the local state of the component as well.

thiyagaraj-eightfold: I understand the use case but there should be a clear distinction if the tab is controlled locally or by the parent. 
The better pattern is to decouple the state in a useTabs hook, so the component becomes more "predictable".

src/components/Tabs/Tabs.context.tsx
```diff
@@ -12,6 +12,10 @@ const TabsProvider = ({ children, onChange, activeTab }: TabsContextProps) => {
     const [currentActiveTab, setCurrentActiveTab] =
         useState<TabValue>(activeTab);
 
+    useEffect(() => {
+        setCurrentActiveTab(activeTab);
```
dkilgore-eightfold, line 16: Reads a bit like Python, but I get it.

dkilgore-eightfold (approved):

PR #106, opened 2022-05-04
feat: abstract out common component props as oc base props, add ability to forward ref

ychhabra-eightfold:


src/components/OcBase/Base.types.ts
```diff
@@ -0,0 +1,6 @@
+import React, { HTMLAttributes } from 'react';
+
+export interface OcBaseProps<T> extends HTMLAttributes<T> {
+    classNames?: string;
```
dkilgore-eightfold, line 4: Should this also contain `style`?

dkilgore-eightfold (changes requested):

src/components/OcBase/Base.types.ts
ychhabra-eightfold: style is already part of HTMLAttributes

src/components/OcBase/Base.types.ts
```diff
@@ -0,0 +1,6 @@
+import React, { HTMLAttributes } from 'react';
+
+export interface OcBaseProps<T> extends HTMLAttributes<T> {
```
dkilgore-eightfold, line 3: @ychhabra-eightfold, come to think of it we may need to discuss this further and include @thiyagaraj-eightfold as he owns the task.

dkilgore-eightfold (changes requested):

src/components/OcBase/Base.types.ts
ychhabra-eightfold: yeah completely agree @dkilgore-eightfold, @thiyagaraj-eightfold please share ur proposal. 
Just to be mindfull of timelines i'm currently blocked on this, which is why I wanted to add data-test-id as a prop in every component.

thiyagaraj-eightfold: Hey -it's not really solving the problem of primitives, it's just creating a standard interface for the component. 
Just standardizing the props.
If it's blocking, let's merge it for now to unblock @ychhabra-eightfold

thiyagaraj-eightfold: Hey -it's not really solving the problem of primitives, it's just creating a standard interface for the component. 
Just standardizing the props.
If it's blocking, let's merge it for now to unblock @ychhabra-eightfold

thiyagaraj-eightfold (approved):

src/components/OcBase/Base.types.ts
ychhabra-eightfold: That was never the intention 🙂. I just wanted to add data-test-id everywhere, this was the only sane way of doing it imo. You can still work on creating primitives.

PR #108, opened 2022-05-05
chore: add atom

thiyagaraj-eightfold:
JIRA TASK (Eightfold Employees Only):
ENG-20643


src/components/Atom/Atom.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { ForwardRefExoticComponent, RefAttributes } from 'react';
+import { ArgumentArray, mergeClasses } from '../../shared/utilities';
+import { OcBaseProps } from '../OcBase';
+
+type AtomProps<T = {}, U = React.Ref<any>> = T &
+    OcBaseProps<U> & {
+        of: keyof HTMLElementTagNameMap;
+        classes: ArgumentArray;
+        children?: any;
```
dkilgore-eightfold, line 9: Should this be more strongly typed? Either a React Node, or JSX?

src/components/Button/BaseButton.tsx
```diff
@@ -150,26 +151,30 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
...
 
         return (
             <>
-                <button
+                <Atom<
+                    React.ButtonHTMLAttributes<HTMLButtonElement>,
+                    HTMLButtonElement
+                >
                     {...rest}
+                    of="div"
```
dkilgore-eightfold, line 169: Should this be enumerated? For example this should be `"button"` correct?

src/components/Inputs/TextInput/TextInput.tsx
```diff
@@ -47,17 +48,14 @@ export const TextInput: FC<TextInputProps> = ({
     const [inputId] = useState<string>(uniqueId(id || 'input-'));
     const inputField: HTMLElement = document.getElementById(inputId);
 
-    const iconClassNames: string = mergeClasses([
-        styles.iconWrapper,
-        styles.leftIcon,
-    ]);
+    const iconClasses = [styles.iconWrapper, styles.leftIcon];
```
dkilgore-eightfold, line 51: nit: type ...

src/components/MatchScore/MatchScore.tsx
```diff
@@ -13,19 +13,16 @@ export const MatchScore: FC<MatchScoreProps> = React.forwardRef(
...
+        ref
     ) => {
         const absTotal: number = Math.abs(total);
-        const matchScoreClasses = mergeClasses(
-            classNames,
-            styles.matchScoreContainer
-        );
         return (
-            <div
+            <Atom<{}, HTMLDivElement>
```
dkilgore-eightfold, line 20: Why is this an empty object?

dkilgore-eightfold (changes requested):

src/components/Button/BaseButton.tsx
ychhabra-eightfold, line 169: The amount of code we need to write increases a lot cause of this change for eg:
`<button>` vs `<Atom<React.ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement> of="button">`

This will both make it cumbersome to use and will effet overall code readability. Also everytime want to now use a div or any element as an atom there will be uncessary js running in the background. 
I think we need to step back and rethink this approach.

ychhabra-eightfold (changes requested):

src/components/Button/BaseButton.tsx
```diff
@@ -150,26 +151,30 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
         );
 
         const getButtonText = (
-            buttonTextClassNames: string,
+            buttonTextClasses: ArgumentArray,
             text: string
         ): JSX.Element => (
-            <span className={buttonTextClassNames}>
+            <Atom of="span" classes={buttonTextClasses}>
```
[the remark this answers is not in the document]
thiyagaraj-eightfold: Yes, that makes sense. But in this case the span is getting a bunch of classes, so thought to atomize it?

src/components/Button/BaseButton.tsx
thiyagaraj-eightfold: @ychhabra-eightfold Yes, there is a level of complexity involved here. 
The tradeoff is to have fine control of what can be passed to a 'div'. Also, atoms are only for primitive components and used internally within the library.

src/components/Atom/Atom.types.ts
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { ArgumentArray } from '../../shared/utilities';
+
+type OcAtomBaseProps<T> = {
```
dkilgore-eightfold, line 4: Update this to use OcBaseProps - perhaps combining this component and those props.

dkilgore-eightfold (changes requested):

dkilgore-eightfold (approved):

src/components/Atom/Atom.types.ts
thiyagaraj-eightfold: This one is slightly different from what is in the codebase, hence I created a new one. Currently, atom base prop accepts a generic and extends it as HTML attributes. Many components have blindly extended spans/divs. This needs to be furthered discussed and iterated.

PR #111, opened 2022-05-10
chore: hooks: adds ut and optimizes match media hook

dkilgore-eightfold:


src/hooks/useMatchMedia.ts
```diff
@@ -16,13 +16,19 @@ export const useMatchMedia = (breakpoint: Breakpoints): boolean => {
     const [threshold, setThreshold] = useState<boolean>(
         window.matchMedia(breakpoint).matches
     );
+    const handleMatchMedia = useCallback(
+        (e: MediaQueryListEvent) => {
+            setThreshold(e.matches);
+        },
+        [threshold]
```
thiyagaraj-eightfold, line 23: threshold passed in the dependency array is not being used. is it intentional? I think an empty [] should work

src/hooks/useMatchMedia.ts
```diff
@@ -16,13 +16,19 @@ export const useMatchMedia = (breakpoint: Breakpoints): boolean => {
...
     useEffect((): void => {
         window
             .matchMedia(breakpoint)
-            .addEventListener('change', (e) => setThreshold(e.matches));
+            .addEventListener('change', (e) => handleMatchMedia(e));
         return window
             .matchMedia(breakpoint)
-            .removeEventListener('change', (e) => setThreshold(e.matches));
-    }, []);
+            .removeEventListener('change', (e) => handleMatchMedia(e));
```
thiyagaraj-eightfold, line 31: ` .removeEventListener('change', handleMatchMedia);`
No need for the anonymous function

thiyagaraj-eightfold (approved):

PR #133, opened 2022-05-24
fix: button: fixes bugs in split and toggle buttons

dkilgore-eightfold:
## SUMMARY:
Fixes bug where split button wrapped unexpectedly and ensures two state button uses badge component that resizes with large numbers

ENG-13981

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull PR and run Octuple by doing `yarn` and then `yarn storybook` from the project root
Resize the window in the Split Button view
Change the number from 8 to 80,000 in the Two State Button counter


src/components/Button/BaseButton.tsx
```diff
@@ -159,10 +159,10 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
         );
 
         return (
-            <>
+            <span className={styles.buttonWrapper}>
```
ychhabra-eightfold, line 162: i'm dont think we should wrap every button in a span, is there any other way we can acheive this ?

ychhabra-eightfold (changes requested):

src/components/Button/BaseButton.tsx
dkilgore-eightfold: Nope

dkilgore-eightfold: spans have no mass, shouldn't have any effect on layout.

PR #168, opened 2022-06-06
fix: handle onScrollLock cleanly, dialog to render inside parent element

ychhabra-eightfold:
## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/components/Dialog/BaseDialog/BaseDialog.tsx
+++ b/src/components/Dialog/BaseDialog/BaseDialog.tsx
@@ -37,7 +37,7 @@ export const BaseDialog: FC<BaseDialogProps> = React.forwardRef(
     ) => {
         const labelId = uniqueId('dialog-label-');
 
-        const { lockScroll, unlockScroll } = useScrollLock(parent);
+        useScrollLock(parent, visible);
 
         const dialogBackdropClasses: string = mergeClasses([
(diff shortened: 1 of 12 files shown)


pbahukhandi-eightfold (approved):

src/components/Dialog/BaseDialog/BaseDialog.tsx
```diff
@@ -31,13 +31,15 @@ export const BaseDialog: FC<BaseDialogProps> = React.forwardRef(
             actionsClassNames,
             dialogWrapperClassNames,
             dialogClassNames,
+            positionStrategy = 'absolute',
+            style,
```
darora-eightfold, line 35: It was coming fixed by default when we tested, changing it to absolute by default won't affect where it is already used?

darora-eightfold (approved):
LGTM

src/components/Dialog/BaseDialog/BaseDialog.tsx
ychhabra-eightfold: Since by default the element is portaled to the body, position absolute will also work. But in cases it doesnt people can change positionStrategy

PR #185, opened 2022-06-15
chore: reorganize utilities, add visuallyHidden util

jhoward-eightfold:
## SUMMARY:
Reorganized utilities by moving to a subdirectory of shared, splitting into separate files per utility, and adding a bundle file which means imports throughout the library don't need to be updated. Added an additional visuallyHidden util (borrowed from mui) to support a11y.

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:


src/shared/utilities/hasWindow.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Utility to verify if the window object exists
+ * @returns {boolean}
+ */
+export const hasWindow = (): boolean => {
```
dkilgore-eightfold, line 5: Recommend changing this and its refs to `canUseDom`

src/shared/utilities/hasWindow.ts
```diff
@@ -0,0 +1,7 @@
...
+    return typeof window !== 'undefined';
```
dkilgore-eightfold, line 6: update to:

```
return !!(
        typeof window !== 'undefined' &&
        window.document &&
        window.document.createElement
    );
};
```

dkilgore-eightfold (changes requested):

PR #187, opened 2022-06-15
feat: button: adds counter badge support to all buttons

dkilgore-eightfold:
## SUMMARY:
Adds counter badge to button

ENG-13961

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [X] Feature Pull Request

## TEST PLAN:
Pull the changes and verify the counter exists in the relevant stories:

Button
> Counter
> Toggle With Counter
> Split With Counter

### DIFF
--- a/src/components/Button/BaseButton.tsx
+++ b/src/components/Button/BaseButton.tsx
@@ -10,6 +10,7 @@ import {
     SplitButton,
 } from './';
 import { Icon, IconSize } from '../Icon';
+import { Badge } from '../Badge';
 import { Breakpoints, useMatchMedia } from '../../hooks/useMatchMedia';
 import { mergeClasses } from '../../shared/utilities';
 
@@ -25,6 +26,7 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
             buttonWidth = ButtonWidth.fitContent,
             checked = false,
             classNames,
+            counter,
             disabled = false,
             disruptive = false,
             dropShadow = false,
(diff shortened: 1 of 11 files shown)


src/components/Button/BaseButton.tsx
```diff
@@ -120,6 +123,11 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
             { [styles.buttonText3]: size === ButtonSize.Small },
         ]);
 
+        const badgeClassNames: string = mergeClasses([
+            styles.counter,
+            buttonTextClassNames,
```
ychhabra-eightfold, line 128: Not sure why this is needed. Whats the reason for this ?

src/components/Button/button.module.scss
```diff
@@ -155,15 +155,39 @@
...
+            background-color: var(--button-counter-hover-background-color);
+        }
+    }
+
+    &:active {
+        .counter {
+            background-color: var(--button-counter-active-background-color);
+        }
+    }
+
```
ychhabra-eightfold, line 183: instead of overriding badge styles, can we use the badge active state ?
and disruptive for diruptive button

src/components/Button/BaseButton.tsx
dkilgore-eightfold: The badge layout requires flex support and to inherit the font size from button.

src/components/Button/Button.types.ts
```diff
@@ -118,6 +114,10 @@ export interface ButtonProps extends NativeButtonProps {
      * The button class names.
      */
     classNames?: string;
+    /**
+     * The button counter number.
+     */
+    counter?: number;
```
ychhabra-eightfold, line 120: can we change this to type string for large number use cases. For eg: we might `3k` instead of `3000`

ychhabra-eightfold (changes requested):

ychhabra-eightfold (approved):

PR #200, opened 2022-06-22
chore: added support for avatar hashing function

ayadav-eightfold:
chore: added support for avatar hashing function


pbahukhandi-eightfold (approved):

src/components/Avatar/Avatar.tsx
```diff
@@ -18,11 +18,16 @@ export const AVATAR_CLASS_SET = [
...
-            () => Math.floor(Math.random() * 100) % AVATAR_CLASS_SET.length,
-            []
-        );
+    (
+        { children, classNames, style, hashingFunction },
+        ref: Ref<HTMLDivElement>
+    ) => {
+        const colorSetIndex = useMemo(() => {
+            if (hashingFunction) {
+                return hashingFunction();
```
pbahukhandi-eightfold, line 27: @ychhabra-eightfold @dkilgore-eightfold : to review

dkilgore-eightfold (approved):

src/components/Avatar/Avatar.tsx
```diff
@@ -18,11 +18,16 @@ export const AVATAR_CLASS_SET = [
 ];
```
ayadav-eightfold, line 18: expose theme set as "theme": type : OcThemeNames(src/components/ConfigProvider/Theming/Theming.types.ts)
prop randomiseTheme: if true, use Math.floor(Math.random() * 100) % AVATAR_CLASS_SET.length;
passing the hashing function as prop

src/components/Avatar/Avatar.tsx
```diff
@@ -65,6 +76,9 @@ export const Avatar: FC<AvatarProps> = React.forwardRef(
             fontSize = '18px',
             iconProps,
             children,
+            hashingFunction,
+            theme,
```
ychhabra-eightfold, line 84: can we make default blue

src/components/Avatar/Avatar.tsx
```diff
@@ -18,16 +19,26 @@ export const AVATAR_CLASS_SET = [
...
+                return Math.floor(hashingFunction()) % AVATAR_THEME_SET.length;
+            }
+            return 0;
+        }, []);
 
         const avatarClasses: string = mergeClasses([
             styles.wrapperStyle,
             classNames,
-            AVATAR_CLASS_SET[colorSetIndex],
+            styles[theme] || AVATAR_THEME_SET[colorSetIndex],
```
ychhabra-eightfold: dont think this will work `styles[theme]` can you confirm ?

src/components/Avatar/Avatar.tsx
```diff
@@ -5,33 +5,48 @@ import styles from './avatar.module.scss';
...
+
         return (
-            <div ref={ref} className={avatarClasses} style={style}>
+            <div
+                ref={ref}
+                className={avatarClasses}
+                style={{
+                    backgroundColor: `var(--${theme}-color-20)`,
+                    color: `var(--${theme}-color)`,
+                    border: `solid 1px var(--${theme}-color-30)`,
```
ychhabra-eightfold, line 46: can we just create a theme to stype map and use that instead of this

ychhabra-eightfold (changes requested):

src/components/Avatar/Avatar.tsx
ayadav-eightfold: If we keep a default theme, we'll never be able to use the hashing function on the basis of priority order we had decided

PR #202, opened 2022-06-23
fix: remove buttons from menu item

ychhabra-eightfold:
## SUMMARY:
fix: remove buttons from menu item

## CHANGE TYPE:

-   [x] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
skip

### DIFF
--- a/src/components/Menu/Menu.types.ts
+++ b/src/components/Menu/Menu.types.ts
@@ -1,13 +1,5 @@
-import { ButtonProps } from '../Button';
 import { ListProps } from '../List';
-
-export interface MenuItem
-    extends Omit<ButtonProps, 'disruptive' | 'onClick' | 'value'> {
-    /**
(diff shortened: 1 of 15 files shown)


src/components/Menu/Menu.types.ts
```diff
@@ -1,27 +1,30 @@
...
     neutral = 'neutral',
 }
 
+export enum MenuSize {
+    large = 'large',
+    medium = 'medium',
+    small = 'small',
+}
+
 export interface MenuProps
```
dkilgore-eightfold, line 16: Should there be a `multiple?: boolean` prop? When this prop, or will the MenuItem input type passed to the component determine this functionality? Would be good to see an updated story to get a better idea.

dkilgore-eightfold (approved):

PR #205, opened 2022-06-23
fix: button: removes focus outline and border style when not in keyboard mode

dkilgore-eightfold:
## SUMMARY:
The `:focus` selector property is not needed. It was reported that the buttons appeared to be stuck in an active state after click. This change removes `:focus` and defers to `.focus-visible:focus-visible`, only showing the focused state when using keyboard via the ConfigProvider.

ENG-22620

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:
NA

## TEST PLAN:
Pull the PR changes, build storybook and verify focus styles and that they only appear when using keyboard in the ConfigProvider story view.

### DIFF
--- a/src/components/Button/button.module.scss
+++ b/src/components/Button/button.module.scss
@@ -6,7 +6,6 @@
     --button-visited-foreground-color: var(--primary-color-80);
     --button-default-outline-color: var(--primary-color);
     --button-hover-outline-color: var(--primary-color-60);
-    --button-focus-outline-color: var(--primary-color-80);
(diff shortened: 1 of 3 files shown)


jhoward-eightfold (approved):
is this the general approach to focus highlighting we should be using?

dkilgore-eightfold:
> is this the general approach to focus highlighting we should be using?

Overall, yes - but there may be some slight differences with where the border shows depending on the component (offset inner or outer)

PR #212, opened 2022-06-29
feat: eng-22373: enable floating buttons

yzhang-eightfold:
## SUMMARY:
Adding floatingButtonProps in BaseButton to enable it floating in botton right corner

https://eightfoldai.atlassian.net/browse/ENG-22373

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [x] Feature Pull Request

## TEST PLAN:
In story book there should be an option to make button floating
image

also try "round" shape and different size, should always work
image

### DIFF
--- a/src/components/Button/BaseButton.tsx
+++ b/src/components/Button/BaseButton.tsx
@@ -40,6 +40,7 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
             split,
             splitButtonChecked = false,
             splitButtonProps,
+            floatingButtonProps,
             style,
             text,
             theme,
@@ -82,6 +83,7 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
             { [styles.pillShape]: shape === ButtonShape.Pill },
             { [styles.dropShadow]: dropShadow },
             { [styles.dark]: theme === ButtonTheme.dark },
+            { [styles.floating]: floatingButtonProps?.enabled },
         ]);
 
         const buttonBaseClassNames: string = mergeClasses([
--- a/src/components/Button/BaseButton.tsx
+++ b/src/components/Button/BaseButton.tsx
@@ -30,6 +30,7 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
             disabled = false,
             disruptive = false,
             dropShadow = false,
+            floatingButtonProps,
             htmlType,
             iconProps,
             id,
@@ -40,7 +41,6 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
             split,
             splitButtonChecked = false,
             splitButtonProps,
-            floatingButtonProps,
             style,
             text,
             theme,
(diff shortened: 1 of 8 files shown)


src/components/Button/button.module.scss
```diff
@@ -163,6 +163,13 @@
         width: fit-content;
     }
 
+    &.floating {
+        position: fixed;
+        bottom: 40px;
+        right: 40px;
+        z-index: $z-index-400;
```
dkilgore-eightfold, line 170: Override both the default and pill shapes here, making the button round when it's icon only. All other scenarios may keep the button shapes 'as-is'. That way we don't have that weird oblong shape.

dkilgore-eightfold (changes requested):

src/components/Button/button.module.scss
yzhang-eightfold: It might help to have traditional button shape floating, I'll add a "round" shape in button.types

src/components/Button/BaseButton.tsx
```diff
@@ -40,6 +40,7 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
             split,
             splitButtonChecked = false,
             splitButtonProps,
+            floatingButtonProps,
```
dkilgore-eightfold, line 43: You'll need to ensure the new `floatingButtonProps` prop is passed down through `PrimaryButton`, `SecondaryButton`, `DefaultButton`, and `NeutralButton`.

dkilgore-eightfold: `SplitButton` and `TwoStateButton` don't need this prop so those may be left 'as-is'.

src/components/Button/BaseButton.tsx
```diff
@@ -80,8 +81,10 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
             { [styles.buttonSize2]: size === ButtonSize.Medium },
             { [styles.buttonSize3]: size === ButtonSize.Small },
             { [styles.pillShape]: shape === ButtonShape.Pill },
+            { [styles.roundShape]: shape === ButtonShape.Round },
```
dkilgore-eightfold, line 84: You'll need to handle this so that when `split` is true, and that `TwoStateButton` may not have a round shape. Those would look funky.

dkilgore-eightfold (changes requested):

src/components/Button/BaseButton.tsx
dkilgore-eightfold: Let's restrict round to buttons with icons only for now.

src/components/Button/BaseButton.tsx
```diff
@@ -80,8 +81,10 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
             { [styles.buttonSize2]: size === ButtonSize.Medium },
             { [styles.buttonSize3]: size === ButtonSize.Small },
             { [styles.pillShape]: shape === ButtonShape.Pill },
+            { [styles.roundShape]: shape === ButtonShape.Round && !split },
```
yzhang-eightfold, line 84: @dkilgore-eightfold Looks like two-state button doesn't call baseButton at all and everything is handled there on its own. So here I'm only handling "split". Verified that when set to round, split button still show rectangular

src/components/Button/BaseButton.tsx
```diff
@@ -80,8 +81,13 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
             { [styles.buttonSize2]: size === ButtonSize.Medium },
             { [styles.buttonSize3]: size === ButtonSize.Small },
             { [styles.pillShape]: shape === ButtonShape.Pill },
+            {
+                [styles.roundShape]:
+                    shape === ButtonShape.Round && !split && !textExists,
```
yzhang-eightfold, line 86: @dkilgore-eightfold Looks like two-state button doesn't call baseButton at all and everything is handled there on its own. So here I'm only handling "split". Verified that when set to round, split button still show rectangular

dkilgore-eightfold (approved):

PR #222, opened 2022-07-07
chore: scoped CSS to octuple modules

ayadav-eightfold:
fix: scoped CSS to octuple modules

### DIFF
--- a/src/components/Inputs/input.module.scss
+++ b/src/components/Inputs/input.module.scss
@@ -2,6 +2,17 @@
     position: relative;
     width: fit-content;
 
+    input[type='submit'],
+    input[type='reset'] {
+        background: none;
+        color: inherit;
+        border: none;
+        padding: 0;
+        font: inherit;
+        cursor: pointer;
+        outline: inherit;
+    }
+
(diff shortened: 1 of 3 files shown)


src/components/Inputs/input.module.scss
```diff
@@ -2,6 +2,17 @@
     position: relative;
     width: fit-content;
 
+    input[type='submit'],
```
ychhabra-eightfold, line 5: this needs to be in search box as well ??

ychhabra-eightfold (changes requested):

src/components/Inputs/input.module.scss
dkilgore-eightfold: What's the scope?

src/styles/base/_css-reset.scss
```diff
@@ -9,8 +9,8 @@
  */
```
dkilgore-eightfold, line 9: We should make this a mixin and include it in all component css. That should do the trick.

dkilgore-eightfold (changes requested):

PR #232, opened 2022-07-11
fix: prevent slider from firing onChange during init

jhoward-eightfold:
## SUMMARY:
There were several setState calls in my layout logic which were triggering re-render. Eliminated these so that layout changes were done directly via ref updates. Also added ResizeObserver for triggering layout updates.

https://eightfoldai.atlassian.net/browse/ENG-23989

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Added a unit test to ensure onChange doesn't fire on initial render.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
         "details": "webpack --config webpack.dev.js --mode development --stats-error-details",
         "analyze": "webpack --config webpack.dev.js --mode production --json > stats.json",
         "prepublish": "yarn build",
-        "test": "jest",
+        "test": "jest --collectCoverage=true",
         "test:update": "jest -u",
         "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
         "format": "prettier --write \"**/*.+(js|jsx|ts|tsx|json|css|scss|md)\"",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
         "details": "webpack --config webpack.dev.js --mode development --stats-error-details",
         "analyze": "webpack --config webpack.dev.js --mode production --json > stats.json",
         "prepublish": "yarn build",
-        "test": "jest",
+        "test": "jest --collectCoverage=true",
```
dkilgore-eightfold, line 18: This may significantly slow down tests.

dkilgore-eightfold (approved):

package.json
jhoward-eightfold: This was already enabled via the previously true config version on line 245. Adding it here and disabling it in the below config results in the coverage still being reported when running the entire test suite via `yarn run test`, but no output when running a single test within vscode or command line running jest directly.

PR #239, opened 2022-07-18
feat: input: add inline prop

dkilgore-eightfold:
## SUMMARY:
Adds inline prop and styling, fixes some bugs with underline.

ENG-2792

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [X] Feature Pull Request

## TEST COVERAGE:
TODO: Add UTs for Inputs

## TEST PLAN:
Pull the PR changes, `yarn` and `yarn storybook` and verify the following props behave as expected:

- inline
- underline
- inline + undeline


src/components/DateTimePicker/Internal/ocpicker.module.scss
```diff
@@ -924,6 +929,11 @@
         &-placeholder {
             input {
                 color: $picker-input-placeholder-color;
+
+                &-disabled,
+                &[disabled] {
+                    @include placeholder($picker-input-foreground-color);
```
jhoward-eightfold, line 935: are we doing anything else visually for disabled input? if its just this it seems insufficient as it will look identical to an empty input

src/components/Inputs/input.module.scss
```diff
@@ -489,6 +451,12 @@
             }
         }
 
+        &.inline {
+            display: flex;
+            flex-direction: column;
+            flex-grow: 1;
+        }
```
jhoward-eightfold, line 458: I'm guessing at some point someone is going to want to stack inline elements on top of each other and have their inputs line up. Give that some thought but try not to lose sleep over it. I'm assuming for now the use case here is above tables, etc. places where we want the vertical spacing minimized

jhoward-eightfold (approved):

src/components/DateTimePicker/Internal/ocpicker.module.scss
dkilgore-eightfold: The opacity dips to 0.3

jhoward-eightfold: Roger.

PR #240, opened 2022-07-18
fix: select: Move chevron indicator to the right side

dkilgore-eightfold:
## SUMMARY:
- adds select shapes and sizes mapped to input
- refines left and right icon layouts in input component

https://user-images.githubusercontent.com/99700808/179634323-e085f701-db88-4ae0-acdc-32cbcde635b7.mp4

ENG-22505

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:
TODO: Add UTs for Select and Input

## TEST PLAN:
Pull latest and run `yarn` and `yarn storybook`. Verify the following:

Select:
shape
size
Icon should now be on the right and correctly reflect the chevron in Figma Theme 1.

### DIFF
--- a/src/components/Inputs/Input.types.ts
+++ b/src/components/Inputs/Input.types.ts
@@ -5,6 +5,11 @@ import { LabelProps } from '../Label';
 import { TooltipTheme } from '../Tooltip';
 import { OcBaseProps } from '../OcBase';
 
+export enum TextInputIconAlign {
+    Left = 'left',
+    Right = 'right',
+}
+
 export enum TextInputSize {
     Flex = 'flex',
     Large = 'large',
@@ -147,7 +152,7 @@ export interface SearchBoxProps
 export interface TextAreaProps
     extends Omit<
         InputProps<HTMLTextAreaElement>,
-        'clearButtonAriaLabel' | 'iconProps' | 'iconButtonProps'
+        'clearButtonAriaLabel' | 'iconProps' | 'iconButtonProps' | 'alignIcon'
     > {
     /**
      * The text area is expandable.
@@ -176,12 +181,6 @@ export interface TextAreaProps
 }
 
 export interface TextInputProps extends InputProps<HTMLInputElement> {
-    /**
-     * option to show the clear input button.
-     * default is true for backward compatibility
-     * @default true
-     */
-    clearable?: boolean;
     /**
      * The input html type.
      * @default 'text'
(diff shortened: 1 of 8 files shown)


src/components/Inputs/Input.types.ts
```diff
@@ -212,6 +211,17 @@ export interface InputProps<T>
...
         'onChange' | 'onFocus' | 'onBlur' | 'onKeyDown'
     > {
+    /**
+     * The input icon alignment.
+     * @default TextInputIconAlign.Left
+     */
+    alignIcon?: TextInputIconAlign;
+    /**
+     * option to show the clear input button.
+     * default is true for backward compatibility
```
jhoward-eightfold, line 221: how many implementations need this? feel like we are early enough in development that we just break it backwards compatibility in this case unless this is desirable in more cases than not.

src/components/Inputs/Input.types.ts
```diff
@@ -212,6 +211,17 @@ export interface InputProps<T>
         OcBaseProps<T>,
         'onChange' | 'onFocus' | 'onBlur' | 'onKeyDown'
     > {
+    /**
+     * The input icon alignment.
+     * @default TextInputIconAlign.Left
```
jhoward-eightfold, line 216: I'm assuming this is the chevron... it looks so much better (and normal) on the right. UX wants the chevron on the left?

src/components/Inputs/input.module.scss
```diff
@@ -252,36 +500,118 @@
...
-            }
+                &.left-icon {
+                    padding: $space-xs calc(#{$space-l} + 4px) $space-xs
+                        $space-xl;
+                }
 
-            &.with-icon-button {
-                padding: $space-xs calc(#{$space-l} + 4px) $space-xs $space-xl;
+                &.right-icon {
+                    padding: $space-xs $space-xl $space-xs
```
jhoward-eightfold, line 514: guessing this is the large variant? hard to tell with number based class names. large variant seemed like the chevron was too far away from the right edge of the select.

jhoward-eightfold (approved):

src/components/Inputs/Input.types.ts
dkilgore-eightfold: UX want the chevron on the right.

src/components/Inputs/input.module.scss
dkilgore-eightfold: will have a look

src/components/Inputs/Input.types.ts
dkilgore-eightfold: Moving clearable to the root interface should be safe.

src/components/Inputs/Input.types.ts
jhoward-eightfold: Agree with that. Looked very funny on the left. Can we make right that the default then? ... or is this icon not the chevron?

PR #258, opened 2022-07-25
chore: lfs file rewrite

dkilgore-eightfold:
## SUMMARY:
The size of the `*.snap` files located in the commit history of the Octuple repo are quite large, resulting in the repo reaching its storage quota before the end of the month whenever `git lfs` is not available. This change uses `git filter-repo` to remove those large files by type from the repo history using the following command:

`git filter-repo --force --path-glob '*.snap' --invert-paths`

Without this change, upon cloning, committing, pushing or pulling developers will encounter the following `git lfs` error when there is no available data plan:

`batch response: This repository is over its data quota. Account responsible for LFS bandwidth should purchase more data packs to restore access.`

Once this change is merged into `main`, then `main-next` and `v1.x-hotfix` should also take this change. Upon completing this task we may then follow steps to `git lfs uninstall` for the main repo ahead of merging `main-next` into `main` .

References:

[newren/git-filter-repo](https://github.com/newren/git-filter-repo)

[How to Remove a Large File from Commit History in Git?](https://tyricwei.medium.com/how-to-delete-large-files-from-git-history-69826a457e74)

[Git - Remove All of a Certain Type of File from the Repository](https://stackoverflow.com/questions/38880436/git-remove-all-of-a-certain-type-of-file-from-the-repository)


jhoward-eightfold (approved):
Just going on record that I am NOT reviewing 600,000+ lines of code. Trusting that what you are doing here is appropriate 😆

PR #275, opened 2022-07-29
fix: selectors: adds size prop, fixes styles to match the latest

dkilgore-eightfold:
## SUMMARY:

- Updated the button CSS
- Updated the Dialog Button type
- Updated the Dialog CSS
- Added props to selectors and selector groups
- updated selector CSS

https://user-images.githubusercontent.com/99700808/181661384-0437c09b-8ef6-42b6-bcec-fff7ddb49658.mp4

https://user-images.githubusercontent.com/99700808/181661407-9629210c-f389-4d58-bd47-47ab841b4777.mp4

ENG-23846: [Button] [Theme2] Default style has the wrong color in pressed state
ENG-23847: [Button] [Theme2] Split style's divider line is not passing accessibility
ENG-23852: [Dialog] [Theme2] cancel button should be neutral version and not default version
ENG-23853: [Modal] [Theme2] Close button should be in neutral style
ENG-23858; [Info Bar] [Theme2] surrounding padding should be 20px. 
ENG-25252: Add ability to disable radio and checkbox groups at parent level
ENG-25728: [Selectors] RadioButton and CheckBox need size prop
ENG-25730: [Selectors] AllowDisabledFocus prop is missing

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull this pr and do `yarn` and `yarn storybook` verify the changes tot he following stories:

RadioButton
CheckBox
Default Button
Split Button
Dialog
Modal (inherits from Dialog)
InfoBar

### DIFF
--- a/src/components/CheckBox/CheckBox.tsx
+++ b/src/components/CheckBox/CheckBox.tsx
@@ -99,10 +99,10 @@ export const CheckBox: FC<CheckboxProps> = React.forwardRef(
             >
                 <input
                     ref={ref}
-                    aria-disabled={allowDisabledFocus}
+                    aria-disabled={disabled}
                     aria-label={ariaLabel}
                     checked={isChecked}
-                    disabled={disabled}
+                    disabled={!allowDisabledFocus && disabled}
                     id={checkBoxId.current}
                     onChange={!allowDisabledFocus ? toggleChecked : null}
                     name={name}
--- a/src/components/CheckBox/CheckBox.tsx
+++ b/src/components/CheckBox/CheckBox.tsx
@@ -1,12 +1,14 @@
 import React, { FC, Ref, useEffect, useRef, useState } from 'react';
 import { generateId, mergeClasses } from '../../shared/utilities';
-import { CheckboxProps, LabelPosition } from './';
+import { CheckboxProps, LabelPosition, SelectorSize } from './';
+import { Breakpoints, useMatchMedia } from '../../hooks/useMatchMedia';
 
 import styles from './checkbox.module.scss';
 
 export const CheckBox: FC<CheckboxProps> = React.forwardRef(
     (
         {
+            allowDisabledFocus = false,
             ariaLabel,
             checked = false,
             classNames,
@@ -17,6 +19,7 @@ export const CheckBox: FC<CheckboxProps> = React.forwardRef(
             labelPosition = LabelPosition.End,
             name,
             onChange,
+            size = SelectorSize.Medium,
             style,
             toggle = false,
             value,
@@ -24,6 +27,11 @@ export const CheckBox: FC<CheckboxProps> = React.forwardRef(
         },
         ref: Ref<HTMLInputElement>
     ) => {
+        const largeScreenActive: boolean = useMatchMedia(Breakpoints.Large);
+        const mediumScreenActive: boolean = useMatchMedia(Breakpoints.Medium);
+        const smallScreenActive: boolean = useMatchMedia(Breakpoints.Small);
+        const xSmallScreenActive: boolean = useMatchMedia(Breakpoints.XSmall);
+
         const checkBoxId = useRef<string>(id || generateId());
         const [isChecked, setIsChecked] = useState<boolean>(
             defaultChecked || checked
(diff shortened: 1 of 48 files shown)


src/components/CheckBox/CheckBox.tsx
```diff
@@ -35,7 +43,27 @@ export const CheckBox: FC<CheckboxProps> = React.forwardRef(
...
         const checkboxWrapperClassNames: string = mergeClasses([
             styles.selector,
+            {
+                [styles.selectorSmall]:
+                    size === SelectorSize.Flex && largeScreenActive,
+            },
+            {
+                [styles.selectorMedium]:
+                    size === SelectorSize.Flex && mediumScreenActive,
+            },
```
jhoward-eightfold, line 53: someone had suggested combining these in a different PR. Not sure if that works. Personally think we should revisit mergeClasses. classNames provides more flexibility:
https://www.npmjs.com/package/classnames

src/components/CheckBox/CheckBox.tsx
```diff
@@ -71,11 +99,12 @@ export const CheckBox: FC<CheckboxProps> = React.forwardRef(
             >
                 <input
                     ref={ref}
+                    aria-disabled={allowDisabledFocus}
```
jhoward-eightfold, line 102: I believe we are doing this wrong. Based on docs, this should mimic the disabled value
 https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled
 
 In which case this might look like such:
 aria-disabled={disabled}
 disabled={!allowDisabledFocus && disabled}

and then styling for disabled state would be based on the value of the aria-disabled attribute... thoughts?

src/components/CheckBox/checkbox.module.scss
```diff
@@ -242,14 +242,237 @@
...
+                margin-left: $space-xxs;
+            }
+
+            &-start {
+                margin-right: $space-xxs;
+            }
+        }
+    }
+
+    &.disabled {
```
jhoward-eightfold, line 429: this is alot of styles for disabled state. is it possible to get savings from doing a not trick on the default hovers, etc?
input:not(.disabled):hover {
  color rules
  }

src/components/InfoBar/infoBar.module.scss
```diff
@@ -3,7 +3,7 @@
     --bg-color: var(--grey-color-70);
     display: flex;
     border-radius: $corner-radius-xl;
-    padding: $space-m;
+    padding: $space-l-fitted;
```
jhoward-eightfold, line 6: what is fitted?

src/components/RadioButton/radio.module.scss
```diff
@@ -122,14 +129,156 @@
...
+    &-large {
+        input {
+            & + label {
+                .radio-button {
+                    height: $radio-large-height;
+                    width: $radio-large-width;
+
+                    &:after {
+                        top: $radio-large-after-top;
+                        left: $radio-large-after-left;
```
jhoward-eightfold, line 152: might be able to save us from some of this repeated precision with more generic styling on the base case
top: 50%;
margin-top: calc(-50%);
left: ?
margin-left: ?

jhoward-eightfold (changes requested):

src/components/CheckBox/CheckBox.tsx
dkilgore-eightfold: Will update.

src/components/InfoBar/infoBar.module.scss
dkilgore-eightfold: No context. we can change the name and update references.

src/components/RadioButton/radio.module.scss
dkilgore-eightfold: That approach has cross-browser pitfalls.

src/components/CheckBox/CheckBox.tsx
dkilgore-eightfold: Ya, this was an idea for the Atom component.

jhoward-eightfold (approved):
thanks for the changes

PR #277, opened 2022-07-29
fix: snackbar: increase snackbar z-index to be more than surface elements

ychhabra-eightfold:
## SUMMARY:
fix: snackbar: increase snackbar z-index to be more than surface elements

https://eightfoldai.atlassian.net/browse/ENG-22772

## CHANGE TYPE:

-   [x] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/styles/themes/_definitions-light.scss
+++ b/src/styles/themes/_definitions-light.scss
@@ -237,7 +237,8 @@ $disabled-zero: 0;
 
 $z-index-auto: auto;
 $z-index-top: 99999; //this is a super weapon. use it wisely.
-$z-index-500: 500; //warning and alerts. needs to be on top of all the components
+$z-index-600: 600; //warning and alerts. needs to be on top of all the components
+$z-index-500: 500;
 $z-index-400: 400;
 $z-index-300: 300;
 $z-index-200: 200;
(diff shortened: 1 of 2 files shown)


src/styles/themes/_definitions-light.scss
```diff
@@ -237,7 +237,8 @@ $disabled-zero: 0;
 
 $z-index-auto: auto;
 $z-index-top: 99999; //this is a super weapon. use it wisely.
-$z-index-500: 500; //warning and alerts. needs to be on top of all the components
+$z-index-600: 600; //warning and alerts. needs to be on top of all the components
```
jhoward-eightfold, line 240: something like this might be a bit too aggressive, but it could ensure we are opening modals, snackbars, popovers, etc at the highest level on the page:
https://bobbyhadz.com/blog/javascript-find-highest-z-index-on-page

jhoward-eightfold (approved):

src/styles/themes/_definitions-light.scss
jhoward-eightfold: Example case: We want snackbars to show up above whatevers on the page, but if someone then opens a modal after that snackbar, we would want that modal to open above the snackbar probably. And if we then fire a snackbar from the modal code, we want that snackbar above the modal.

ychhabra-eightfold: This seems like a nice approach, but we might need experiment with it to see how well it works.

PR #279, opened 2022-08-01
fix: radiobutton: fixes width of radio element in flex when layout is constrained

dkilgore-eightfold:
## SUMMARY:
When the RadioButton is used in a thinner layout, the first element gets distorted. This change ensures the width is consistent.

ENG-25650

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Verify the additional stories by either pulling the pr and doing `yarn`, and `yarn storybook` or implement a radio group using the pr CodeSandbox React build, resize the browser to test the layout of the group.

### DIFF
--- a/src/components/RadioButton/radio.module.scss
+++ b/src/components/RadioButton/radio.module.scss
@@ -23,13 +23,14 @@
 
         & + label {
             .radio-button {
-                position: relative;
-                height: $radio-medium-height;
-                width: $radio-medium-width;
-                border-radius: 50%;
                 border: $space-xxxs solid var(--grey-color-70);
+                border-radius: 50%;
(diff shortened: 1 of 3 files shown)


src/components/RadioButton/radio.module.scss
```diff
@@ -23,13 +23,14 @@
 
         & + label {
             .radio-button {
```
jhoward-eightfold, line 25: We should work to avoid deeply nesting css selectors when possibly to reduce specificity and css output

jhoward-eightfold (approved):

PR #283, opened 2022-08-02
fix: css: scope normalize css reset to octuple components

dkilgore-eightfold:
## SUMMARY:
Scopes normalize.css reset to components, keeps 2 low-risk, high ROI global resets.

Prior to this change https://github.com/necolas/normalize.css was being applied globally. This may contribute to css collisions within host apps. Scoping these resets will ensure only Octuple components will benefit.

Also adds vertical margins to menu items.

ENG-26154

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull the pr and do `yarn`, then `yarn install`, verify all components using Storybook.


src/components/Table/Tests/__snapshots__/Table.pagination.renders.shot
```diff
@@ -2209,7 +2209,7 @@ LoadedCheerio {
                           "aria-checked": "true",
                           "aria-disabled": "false",
                           "aria-pressed": "true",
-                          "class": "pagination-button active button button-neutral button-medium",
+                          "class": "pagination-button active button button-neutral button button-medium",
```
jhoward-eightfold, line 2212: why did the additional and highly generic button class get added? is this desirable? does it get hashed?

src/src/components/PersistentBar/__snapshots__/PersistentBar.stories.storyshot
```diff
@@ -143,7 +143,7 @@ exports[`Storyshots Persistent Bar Bottom Bar With Text 1`] = `
     <button
       aria-disabled={false}
       aria-label="Primary"
-      className="my-primary-btn-class button button-primary pill-shape"
+      className="my-primary-btn-class button button-primary button pill-shape"
```
jhoward-eightfold, line 146: and some of these now have two generic button classes applied

jhoward-eightfold (approved):
disregard my question regarding button class. i see where it came from and that it gets hashed. If you are good with this, carry on.

src/components/Table/Tests/__snapshots__/Table.pagination.renders.shot
dkilgore-eightfold: Needed to add this because the button doesn't have a wrapping element.

PR #290, opened 2022-08-04
feat: adds headerbuttonprops, headericon, bodypadding, and overlay props to surface ui

dkilgore-eightfold:
## SUMMARY:
Some scenarios require the following:

- back button
- no padding
- no overlay

This change adds these props to support the above in Dialog, Modal and Panel

https://user-images.githubusercontent.com/99700808/182916258-6f406ac9-8807-4d70-a08d-85c1c378ffea.mp4

ENG-26194

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [X] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
pull the pr branch and do `yarn`, then `yarn install`, verify the chnages in the Dialog, Modal and Panel stories, or use the CodeSandbox CI url to build your own.


src/components/Dialog/BaseDialog/BaseDialog.tsx
```diff
@@ -18,26 +18,31 @@ export const BaseDialog: FC<BaseDialogProps> = React.forwardRef(
...
             header,
+            headerButtonProps,
             headerClassNames,
-            body,
-            bodyClassNames,
-            actions,
-            actionsClassNames,
-            dialogWrapperClassNames,
-            dialogClassNames,
+            headerIcon = IconName.mdiArrowLeftThick,
```
ychhabra-eightfold, line 34: Can we have no headerIcon by default as this will break existing users functionalty

src/components/Dialog/BaseDialog/BaseDialog.tsx
```diff
@@ -50,10 +55,13 @@ export const BaseDialog: FC<BaseDialogProps> = React.forwardRef(
             styles.dialogBackdrop,
             dialogWrapperClassNames,
             { [styles.visible]: visible },
+            { [styles.modeless]: overlay === false },
+            { [styles.modelessMask]: overlay === false && maskClosable },
```
ychhabra-eightfold, line 59: this can be changed into one single object

src/components/Dialog/BaseDialog/base-dialog.module.scss
```diff
@@ -17,6 +24,10 @@
             justify-content: space-between;
             align-items: center;
 
+            &-button {
+                margin-right: 16px;
```
ychhabra-eightfold, line 28: use variable here

src/components/Panel/Panel.tsx
```diff
@@ -28,30 +28,34 @@ export const Panel = React.forwardRef<PanelRef, PanelProps>(
...
+            bodyPadding = true,
             children,
-            placement = 'right',
+            closable = true,
+            closeButtonProps,
+            closeIcon = IconName.mdiClose,
+            footerClassNames,
+            headerButtonProps,
+            headerClassNames,
+            headerIcon = IconName.mdiArrowLeftThick,
```
ychhabra-eightfold, line 41: no default header icon

src/components/Panel/panel.module.scss
```diff
@@ -69,6 +75,10 @@
             display: flex;
             justify-content: space-between;
 
+            &-button {
+                margin-right: 16px;
```
ychhabra-eightfold, line 79: use var here

ychhabra-eightfold (changes requested):

src/components/Dialog/BaseDialog/BaseDialog.tsx
dkilgore-eightfold: unfortunately, maskClosable is already part of the API. need to use 'as is'.

src/components/Dialog/BaseDialog/BaseDialog.tsx
dkilgore-eightfold: Hold-on - this is a new prop - how will it break existing apps?

src/components/Dialog/BaseDialog/BaseDialog.tsx
dkilgore-eightfold: We should deprecate using a decorator, but not remove the prop until the next major release.

src/components/Dialog/BaseDialog/BaseDialog.tsx
ychhabra-eightfold: Cause if we show it by default, it will start showing up in places where the developer had not planned for

src/components/Dialog/BaseDialog/BaseDialog.tsx
ychhabra-eightfold: My comment was to move all the styles into a single object.
```suggestion
            { [styles.modeless]: overlay === false,
              [
              styles.modelessMask]: overlay === false && maskClosable },
```

src/components/Dialog/BaseDialog/BaseDialog.tsx
dkilgore-eightfold: This con doesn't show unless the headerIconProps are defined.

src/components/Panel/Panel.tsx
dkilgore-eightfold: This Icon doesn't show unless the headerIconProps are defined.

ychhabra-eightfold (approved):

PR #294, opened 2022-08-08
chore: configure codesandbox to build storybook

jhoward-eightfold:
## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -2,7 +2,6 @@
     "buildCommand": "build-ci",
     "sandboxes": ["@eightfold.ai/octuple", "/storybook-static"],
     "publishDirectory": {
-        "@eightfold.ai/octuple": "lib",
         "/storybook-static": "storybook-static"
     },
     "node": "16"
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,5 +1,9 @@
 {
-    "buildCommand": "build-storybook",
-    "sandboxes": ["new", "vanilla", "/dist"],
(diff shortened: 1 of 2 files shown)


.codesandbox/ci.json
```diff
@@ -1,4 +1,7 @@
 {
-    "sandboxes": ["new", "vanilla"],
```
dkilgore-eightfold, line 2: Please add this back to fix CodeSandbox CI.

dkilgore-eightfold (changes requested):

.codesandbox/ci.json
dkilgore-eightfold: Will also be needed once we figure out how to get CodeSandbox links in Storybook :)

dkilgore-eightfold (approved):

.codesandbox/ci.json
```diff
@@ -1,4 +1,9 @@
...
+    "buildCommand": "build-ci",
+    "sandboxes": ["@eightfold.ai/octuple", "/storybook-static"],
```
dkilgore-eightfold, line 3: Looks like this config might need to be updated.

dkilgore-eightfold (changes requested):

.codesandbox/ci.json
jhoward-eightfold: Seems to be expecting a package.json in storybook-static. Unfortunately, documentation is on this config is rather light
https://codesandbox.io/docs/ci#configuration

PR #300, opened 2022-08-11
fix: select: fixes select clear when items are toggled

dkilgore-eightfold:
## SUMMARY:
Octuple Select w/ clear component ([link](https://eightfoldai.github.io/octuple.github.io/?path=/story/select--with-clear))

Steps to reproduce the bug:

Select School from the dropdown

You see the clear button (x) to the left of “School”

Select School again.

You don’t see the clear button

Select School again and you’ll see the clear button.

https://user-images.githubusercontent.com/99700808/184056277-b2f07ef3-76a4-4b47-87dc-984d3b36efb8.mp4

ENG-26684

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull the pr changes and run `yarn`, then `yarn storybook`, verify changes to the Select With Clear story, or implement a clearable select via CodeSandbox CI

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useState, useEffect, Ref } from 'react';
+import React, { FC, useEffect, useState, Ref } from 'react';
 
 import { mergeClasses } from '../../shared/utilities';
 import { Dropdown } from '../Dropdown';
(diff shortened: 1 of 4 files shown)


src/components/Select/Select.tsx
```diff
@@ -387,7 +388,7 @@ export const Select: FC<SelectProps> = React.forwardRef(
                             shape={selectShapeToTextInputShapeMap.get(shape)}
                             size={selectSizeToTextInputSizeMap.get(size)}
                             value={
-                                getSelectedOptionText() && !dropdownVisible
+                                !dropdownVisible
                                     ? getSelectedOptionText()
                                     : undefined
```
jhoward-eightfold, line 393: does it matter at all that this changed the output from not including a value to including an empty value? I'm guessing no

jhoward-eightfold (approved):

src/components/Select/Select.tsx
dkilgore-eightfold: This check is only here to handle filterable multiple. We currently hide the input string when dropdown is visible here because it's a real bag of tricks to get the UX right. Planning to remove as soon as I figure out how to show the concatenated string, read from it and then update the selected items when the string is updated - probably using serialization.

PR #301, opened 2022-08-11
fix: button: use configprovider focus visible styles

dkilgore-eightfold:
## SUMMARY:
A recent update to remove some global resets to `<button>` elements regressed the `focus-visible` styles. This change fixes and applies the global style to all buttons in Octuple components

ENG-26733

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
pull the pr branch and run `yarn`, then `yarn install` and verify via the `ConfigProvider` story the styles via tabbing around the components (excluding the color pickers). Or implement a `button` in a `ConfigProvider` using CodeSandbox CI.

### DIFF
--- a/src/components/Button/button.module.scss
+++ b/src/components/Button/button.module.scss
@@ -172,6 +172,12 @@
             background-color: var(--button-counter-active-background-color);
         }
     }
+
+    /** Hides the browser default keyboard focus-visible styles
+        use the ConfigProvider instead */
+    &:focus-visible {
+        outline: none;
+    }
 }
 
 .button-text-large {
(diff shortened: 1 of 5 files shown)


src/components/Button/button.module.scss
```diff
@@ -172,6 +172,12 @@
             background-color: var(--button-counter-active-background-color);
         }
     }
+
+    /** Hides the browser default keyboard focus-visible styles
+        use the ConfigProvider instead */
```
jhoward-eightfold, line 177: nit: this is a separate instruction i think but reads like a run-on sentence

Not sure how it works with css modules, but in my previous codebase sass comments using  "/** */" end up in the outputted css whereas comments using "//" do not.

jhoward-eightfold (approved):

PR #306, opened 2022-08-12
feat: tooltip: add sizes to tooltip

ychhabra-eightfold:
## SUMMARY:
feat: tooltip: add sizes to tooltip

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -1,13 +1,13 @@
 import React, { FC, useEffect, useRef, useState } from 'react';
 import {
-    useFloating,
-    shift,
-    autoUpdate,
     arrow,
+    autoUpdate,
     offset as fOffset,
+    shift,
+    useFloating,
 } from '@floating-ui/react-dom';
 import { FloatingPortal } from '@floating-ui/react-dom-interactions';
-import { TooltipProps, TooltipTheme } from './Tooltip.types';
+import { TooltipProps, TooltipSize, TooltipTheme } from './Tooltip.types';
 import {
     ConditionalWrapper,
     generateId,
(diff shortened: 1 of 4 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,13 +1,13 @@
 import React, { FC, useEffect, useRef, useState } from 'react';
 import {
-    useFloating,
-    shift,
-    autoUpdate,
     arrow,
+    autoUpdate,
     offset as fOffset,
+    shift,
+    useFloating,
```
jhoward-eightfold, line 7: If we want to all get on the same page with alphasorting, etc, there is a vscode plugin called TypeScript Hero which will auto sort import statements.

I believe you were opposed to adding a .vscode configuration file that would recommend extensions? Lets revisit that conversation soon. Its a relatively common thing for open source projects and can make onboarding new contributors easier.

src/components/Tooltip/tooltip.module.scss
```diff
@@ -74,4 +73,16 @@ $tooltip-arrow-shadow: 1px 1px 2px rgba(15, 20, 31, 0.12);
...
+        max-width: 140px;
+    }
+
+    &.medium {
+        max-width: 240px;
+    }
+
+    &.large {
+        max-width: 360px;
+    }
```
jhoward-eightfold, line 87: is there a default, or it will be allowed to grow as large as necessary?

jhoward-eightfold (approved):

src/components/Tooltip/Tooltip.tsx
ychhabra-eightfold: sure we can discuss that, but i dont use vscode 😄

src/components/Tooltip/tooltip.module.scss
ychhabra-eightfold: good catch default should be small, making that change

PR #312, opened 2022-08-18
fix: badge: fix badge font size

ychhabra-eightfold:
## SUMMARY:
fix: badge: fix badge font size

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/components/Badge/badge.module.scss
+++ b/src/components/Badge/badge.module.scss
@@ -1,4 +1,5 @@
 .badge {
+    @include octuple-chrome();
     --bg: var(--grey-color-20);
     --label: var(--grey-color-80);
     background-color: var(--bg);
--- a/src/components/Badge/Badge.tsx
+++ b/src/components/Badge/Badge.tsx
@@ -14,7 +14,7 @@ export const Badge: FC<BadgeProps> = ({
 }) => {
     const badgeClasses: string = mergeClasses([
         styles.badge,
-        'header6',
+        styles.octupleH6,
         { [styles.disruptive]: disruptive },
         classNames,
         { [styles.active]: active },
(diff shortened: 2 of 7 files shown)


src/components/Badge/badge.module.scss
```diff
@@ -1,4 +1,5 @@
 .badge {
+    @include octuple-chrome();
```
jhoward-eightfold, line 2: Does this add the entire chrome css? That would not be good

ychhabra-eightfold: at bundling time everything else should be removed. but i also dont like this approach, with the common styling moved to a mixin i dont see any other approach. Let me know if there is a better way

jhoward-eightfold: I'm also not sure about this approach with the base css. Dylan's main concern is creating unintended negative impact in the existing apps if the base styles are too opinionated. I'm thinking the library shouldn't have to concern itself with that so much. 

in _helpers.scss he has added an .octuple{} wrapper around the base style output... if you can put an octuple class on a parent element, either within the component or within the implementation, it will pick up the base styles. For instance, since TA is going all in on octuple, perhaps an "octuple" class on the body or something?

Alternatively, maybe a sub mixin for the specific header style that octuple-chrome() could call and this implementation point could also call?

@mixin octuple-h6 {
    h6,
    .octuple-h6 {
        font-weight: $text-font-weight-semibold;
        font-size: $text-font-size-2;
        line-height: $text-line-height-2;
        @include content-padding($text-line-height-2);
    }
}

@mixin octuple-chrome {
  ...
  @include octuple-h6();
}

PR #316, opened 2022-08-19
fix: table: remove left border from last fixed cell

ychhabra-eightfold:
## SUMMARY:
fix: table: remove let border from last fixed cell

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/components/Table/Internal/octable.module.scss
+++ b/src/components/Table/Internal/octable.module.scss
@@ -276,6 +276,10 @@
     thead {
         td,
         th {
+            font-weight: $text-font-weight-semibold;
+            font-size: $text-font-size-2;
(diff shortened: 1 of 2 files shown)


src/components/Table/Internal/octable.module.scss
```diff
@@ -171,8 +171,6 @@
 
         &-fix-right-first,
         &-fix-right-last {
-            box-shadow: -1px 0 0 $table-border-color;
```
dkilgore-eightfold, line 174: This should be configurable via props. The actions column in Succession Planning needs a drop shadow. We'll also need to determine whether to only show actions on hover, etc.

PR #326, opened 2022-08-24
feat: add loading state in button and tabs

ychhabra-eightfold:
## SUMMARY:
feat: add loading state in button and tabs

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/components/Loader/Loader.types.ts
+++ b/src/components/Loader/Loader.types.ts
@@ -1,0 +1,11 @@
+import { OcBaseProps } from '../OcBase';
+
+export enum LoaderSize {
+    Large = 'large',
+    Medium = 'medium',
+    Small = 'small',
+}
+
+export interface LoaderProps extends OcBaseProps<HTMLDivElement> {
+    size?: LoaderSize;
+}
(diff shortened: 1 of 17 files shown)


src/components/Loader/Loader.types.ts
```diff
@@ -0,0 +1,11 @@
+import { OcBaseProps } from '../OcBase';
+
+export enum LoaderSize {
```
dkilgore-eightfold, line 3: Suggest using a type here, rather than an enum so a string may be passed in implementation. I'm going to be moving a lot of these enums in this direction. e.g:

`export type LoaderSize = 'flex' | 'large' | 'medium' | 'small' | undefined;`

Adding the flex type and using the `Matchmedia`, `Breakpoint` hooks, etc will ensure this will work with the upcoming Form context with no need for additional mapping.

dkilgore-eightfold (changes requested):

dkilgore-eightfold (approved):

jhoward-eightfold:
lgtm

PR #335, opened 2022-08-31
fix: select: removes portal property from tooltip and replace with position

dkilgore-eightfold:
## SUMMARY:
The Tooltip portal property renders the tooltip below surface UI when surface UI is the parent. Reverting this change and use position prop instead.

ENG-28213

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST PLAN:
pull the PR branch then do `yarn` and `yarn storybook` to verify the tooltip position and it still shows for the typical use case.

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -293,8 +293,8 @@ export const Select: FC<SelectProps> = React.forwardRef(
                     <Tooltip
                         classNames={styles.selectTooltip}
                         content={value.text}
+                        placement={'top'}
                         theme={TooltipTheme.dark}
-                        portal
                     >
                         <Pill
                             ref={(ref) => (pillRefs.current[index] = ref)}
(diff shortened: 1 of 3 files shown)


src/components/Select/Select.tsx
```diff
@@ -293,8 +293,8 @@ export const Select: FC<SelectProps> = React.forwardRef(
                     <Tooltip
                         classNames={styles.selectTooltip}
                         content={value.text}
+                        placement={'top'}
```
jhoward-eightfold, line 296: does our tooltip not support auto placement based on position within the viewport? Seems like you are having to push the storybook example away from the edge to support this, and I would image we might hit scenarios where this will be problematic.

dkilgore-eightfold: It doesn't. Will need to implement this. However, in this scenario the tooltip would be hidden by the dropdown anyways, even with edge detection.

jhoward-eightfold (approved):

PR #349, opened 2022-09-07
fix: adjust peer dependencies

jhoward-eightfold:
## SUMMARY:
When trying to install Octuple or other npm packages, it is necessary to use the --force flag if not on the very specific react version referenced in peerDepedencies provided by Octuple. Relaxing this rule and assuming we should at least support the react version currently in use for eightfold.ai applications.

NA

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     },
     "peerDependencies": {
         "react": ">=16.14.0",
-        "react-dom": "=>16.14.0"
+        "react-dom": ">=16.14.0"
     },
     "devDependencies": {
         "@babel/core": "7.17.5",
--- a/package.json
+++ b/package.json
@@ -71,11 +71,8 @@
         "shallowequal": "1.1.0"
     },
     "peerDependencies": {
-        "@types/react": "17.0.39",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -71,11 +71,8 @@
         "shallowequal": "1.1.0"
     },
     "peerDependencies": {
-        "@types/react": "17.0.39",
-        "@types/react-dom": "17.0.11",
-        "react": "17.0.2",
-        "react-app-polyfill": "3.0.0",
```
jhoward-eightfold, line 77: presuming this isn't actually necessary as the polyfills should get added as needed in compilation

package.json
```diff
@@ -71,11 +71,8 @@
         "shallowequal": "1.1.0"
     },
     "peerDependencies": {
-        "@types/react": "17.0.39",
-        "@types/react-dom": "17.0.11",
```
jhoward-eightfold, line 75: tho nice to have, don't think types are hard dependencies

PR #351, opened 2022-09-08
chore: remove peer dependencies

jhoward-eightfold:
## SUMMARY:
Peer dependencies are too strict and some likely unnecessary. Pegging peerDeps to >=16.8 for react, as this is when hooks were introduced. This is a best guess.

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:


package.json
```diff
@@ -20,7 +20,7 @@
         "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
         "format": "prettier --write \"**/*.+(js|jsx|ts|tsx|json|css|scss|md)\"",
         "chromatic": "chromatic --exit-zero-on-changes",
-        "prepare": "husky install && install-peers",
```
jhoward-eightfold, line 23: we don't need to install peers, as i've moved those that are need to actual dependencies. Uninstalled install-peers and install-peers-cli

package.json
```diff
@@ -64,18 +66,17 @@
...
+        "react-dom": "17.0.2",
         "react-flip-toolkit": "7.0.13",
         "react-is": "18.1.0",
         "resize-observer-polyfill": "1.5.1",
         "scroll-into-view-if-needed": "2.2.25",
         "shallowequal": "1.1.0"
     },
     "peerDependencies": {
-        "@types/react": "17.0.39",
-        "@types/react-dom": "17.0.11",
```
jhoward-eightfold, line 75: Seems safe to remove the types from peer deps? though possibly incorrect types would get passed to the octuple components, causing issues? Thoughts?

package.json
```diff
@@ -64,18 +66,17 @@
...
-        "react": "17.0.2",
-        "react-app-polyfill": "3.0.0",
```
jhoward-eightfold, line 77: This is seemingly only used by jest setupFiles... I've moved it to devDependencies. Its certainly not a peer dependency unless I'm missing something

package.json
```diff
@@ -64,18 +66,17 @@
         "dom-align": "1.12.3",
         "lottie-web": "5.8.1",
         "memoize-one": "6.0.0",
+        "react": "17.0.2",
```
dkilgore-eightfold, line 69: If React is moved into deps, this will be installed in vscode along with its version of React.

package.json
dkilgore-eightfold: Should be okay.

package.json
```diff
@@ -64,18 +66,17 @@
...
         "scroll-into-view-if-needed": "2.2.25",
         "shallowequal": "1.1.0"
     },
     "peerDependencies": {
-        "@types/react": "17.0.39",
-        "@types/react-dom": "17.0.11",
-        "react": "17.0.2",
-        "react-app-polyfill": "3.0.0",
-        "react-dom": "17.0.2"
+        "react": ">=16.8",
```
dkilgore-eightfold, line 78: Why would this be necessary if React is in deps? To trigger a warning at compile time in the host app? Moving React into deps may cause module collisions in the host. @ychhabra-eightfold - would like your take on this too.

package.json
dkilgore-eightfold: Should be okay as a regular dep, as Octuple is already compiled to js in the host.

package.json
```diff
@@ -20,7 +20,7 @@
         "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
```
ychhabra-eightfold, line 20: can we prerelease this in a dev branch, and test out the implications. then we can merge to master

ychhabra-eightfold (changes requested):

package.json
jhoward-eightfold: I see. How do other libraries handle this?
Ant lists react as a devDep and peerDep (https://github.com/ant-design/ant-design/blob/master/package.json)... maybe thats the way to go. Such an approach means you don't need the something like install-peers, but you don't add the overhead of the an additional react version for implementations.

package.json
jhoward-eightfold: It would really be about the warnings, but I think following the model ant uses is probably a better approach.

package.json
jhoward-eightfold: Let me see if I can build and install locally.

ychhabra-eightfold (approved):

PR #356, opened 2022-09-09
fix: search box: icon button default htmltype should be button path should have a default

dkilgore-eightfold:
## SUMMARY:
The search box icon button didn't have any defaults, and the icon button was causing page refreshes due to a missing htmlType, and if customized - the icon path may be undefined.

The InfoBar button types were incorrectly using NeutralButton, this change updates to use SystemUIButton.

Also cleans up some type annotations.

ENG-2959
ENG-23859

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Run `yarn` and `yarn storybook`. Verify the changes to InfoBar and SearchBox stories or use Codesandbox CI to test.


src/components/Inputs/SearchBox/SearchBox.test.tsx
```diff
@@ -22,4 +22,10 @@ describe('SearchBox', () => {
         const wrapper = mount(<SearchBox />);
         expect(wrapper.containsMatchingElement(<SearchBox />)).toEqual(true);
     });
+    test('text input renders with partially customized icon button', () => {
```
jhoward-eightfold, line 25: > partially customized icon button
what does this mean? what is partial about it?

jhoward-eightfold (approved):

src/components/Inputs/SearchBox/SearchBox.test.tsx
dkilgore-eightfold: Like, only 1 property is set. Tests that the defaults work and the component doesn't cause and error.

PR #396, opened 2022-09-28
feat: Enable readOnly for pills without buttons

szhang-eightfold:
## SUMMARY:
Enable readOnly for pills when buttons are present
https://eightfoldai.atlassian.net/jira/software/c/projects/ENG/boards/50?modal=detail&selectedIssue=ENG-30670
## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [x] Feature Pull Request

## TEST COVERAGE:

-   [x] Tests for this change already exist

## TEST PLAN:
- go to storybook's Pill component canvas
- In default Pills and Pills with icon sections
- check that they are not hoverable and no hover effect are implemented
- switch to other sections (with buttons)
- check that they are hoverable and have hover effects


src/components/Pills/Pill.tsx
```diff
@@ -74,13 +75,15 @@ export const Pill: FC<PillProps> = React.forwardRef(
...
             { [styles.tagPillsRtl]: htmlDir === 'rtl' },
+            { [styles.readOnly]: readOnly === true },
         ]);
         return (
             <div
                 {...rest}
                 className={tagClassName}
                 style={{ ...style, color }}
                 ref={ref}
+                tabIndex={readOnly ? -1 : 0}
```
dkilgore-eightfold, line 86: if a element is readOnly it still needs to be accessible, please remove this line and handle any event, as we do with disable.

dkilgore-eightfold (changes requested):

src/components/Pills/Pill.tsx
jhoward-eightfold: Typically only interactive elements are in the tab order. In a readOnly situation the pill/label is informational only and the screen reader can read it, but there's no reason for a user to actually tab to it.
That being said, since this is a div and not an anchor (as it seems to behave by default), it wouldn't be in the tab order anyway. This change is probably an enhancement to that behavior.

However, I'm wondering if it would be wiser for us to make readOnly internal and set it based on the presence of an onClick handler.

Thoughts?

jhoward-eightfold: Looking at the code below, pills are buttons... And this is where the tab order and hover are happening... Why did we decide to make the pill always a button? In some scenarios that makes sense (click to add to a list), but generally I wouldn't expect a pill to be a button

dkilgore-eightfold: read only will need to be readable by accessible tech such as Narrator, VoiceOver, Jaws, etc. Excluding from tab order, or manipulating tab order other than 0 is a bad practice. In this case, the events should be handled when read only, and the UI should reflect the styles @szhang-eightfold applied, but should still be keyboard accessible.

Any clickable element other than inputs should be a button too, or have its role for the same reason. Much like anything that looks like a list should be a `<ul> `or `<ol>`, if we don't address A11y early wherever possible - the cost is tremendous in terms of fixing those kinds of bugs.

dkilgore-eightfold: A good analog implementation of what might be good here would be how we do with allowDisabledFocus in button, accept not with its styles.

dkilgore-eightfold: @szhang-eightfold That being said, only a few tweaks to remove the above and handle those events, I think this change will be good to go.

PR #404, opened 2022-10-02
feat: expandable text input implementation

mfazil-eightfold:
## SUMMARY:
Added expandable search/input box.

image

image

https://eightfoldai.atlassian.net/browse/ENG-30444

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [X] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/components/Inputs/input.module.scss
+++ b/src/components/Inputs/input.module.scss
@@ -2,9 +2,12 @@
     position: relative;
     width: fit-content;
 
+    $input-small-size: 28px;
+    $input-medium-size: 36px;
+    $input-large-size: 44px;
+
     input {
         @include text-overflow;
-
         background-color: var(--background-color);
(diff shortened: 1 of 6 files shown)


src/components/Inputs/input.module.scss
```diff
@@ -1438,6 +1441,109 @@
...
+            width: $input-small-size;
+        }
+
+        &.input-medium .expandable-wrapper {
+            width: $input-medium-size;
+        }
+
+        &.input-large .expandable-wrapper {
+            width: $input-large-size;
+        }
```
jhoward-eightfold, line 1456: Is there a way where we can do this that the implementation dictates the appropriate sizes? It seems too opinionated to hard code widths, even when giving three options.. and particular for the expanded state which will always be 500px below.

PR #405, opened 2022-10-03
fix: a11y fixes in cases of portal dropdown

ktorlakonda-eightfold:
## SUMMARY:
- fixing focus issue in dropdown portal mode

https://eightfoldai.atlassian.net/browse/ENG-30463

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
TBD

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -116,12 +116,17 @@ export const Dropdown: FC<DropdownProps> = React.memo(
                 update
             );
         }, [refs.reference, refs.floating, update]);
-
         useAccessibility(
             trigger,
-            portal ? refs.floating : refs.reference,
+            refs.reference,
             toggle(true),
-            toggle(false)
+            portal
+                ? (e) => {
+                      if (e?.key == 'Enter') {
+                          toggle(false);
+                      }
+                  }
+                : toggle(false)
         );
         const dropdownClasses: string = mergeClasses([
             dropdownClassNames,
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -117,7 +117,12 @@ export const Dropdown: FC<DropdownProps> = React.memo(
             );
         }, [refs.reference, refs.floating, update]);
 
-        useAccessibility(trigger, refs.reference, toggle(true), toggle(false));
+        useAccessibility(
+            trigger,
+            portal ? refs.floating : refs.reference,
+            toggle(true),
+            toggle(false)
+        );
         const dropdownClasses: string = mergeClasses([
             dropdownClassNames,
             styles.dropdownWrapper,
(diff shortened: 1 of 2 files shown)


src/hooks/useAccessibility.ts
```diff
@@ -15,13 +15,15 @@ export const useAccessibility = (
     const triggerEvent: string = ACCESSIBILITY_EVENTS_MAP[trigger];
     const handleAccessibility = useCallback(
         (e) => {
-            if (ref.current?.contains?.(e.target)) {
-                if (triggerEvent != 'keydown' || e.key == 'Enter') {
-                    return onEnter?.();
+            if (ref.current) {
```
dkilgore-eightfold, line 18: The handler is specific, but the hook is generalized. You should create a switch by key using `shared/utilities/eventKeys.ts` so this may be extended without being refactored.

dkilgore-eightfold (changes requested):

src/components/Dropdown/Dropdown.tsx
```diff
@@ -117,7 +117,12 @@ export const Dropdown: FC<DropdownProps> = React.memo(
             );
         }, [refs.reference, refs.floating, update]);
 
-        useAccessibility(trigger, refs.reference, toggle(true), toggle(false));
+        useAccessibility(
```
dkilgore-eightfold, line 120: useAccessibility suggests this hook may be used for other keyboard  scenarios. This hook will need to support more than toggling the visibility of only this component.

dkilgore-eightfold: Suggest stubbing in a switch statement using this scenario as the first use case.

ktorlakonda-eightfold: this PR fixes a11y related issue in case of `portal=true`. 
we can take up more a11y support in the hook but it seems like a out of scope for this PR

src/hooks/useAccessibility.ts
ktorlakonda-eightfold: we are not changing the existing functionality reg. the events.

src/components/Dropdown/Dropdown.tsx
dkilgore-eightfold: You should rename this hook to be more specific. Any refactor would put the scope of your hook at risk in the future.

dkilgore-eightfold (changes requested):

src/components/Dropdown/Dropdown.tsx
```diff
@@ -116,8 +116,18 @@ export const Dropdown: FC<DropdownProps> = React.memo(
                 update
             );
         }, [refs.reference, refs.floating, update]);
-
-        useAccessibility(trigger, refs.reference, toggle(true), toggle(false));
+        useAccessibility(
```
dkilgore-eightfold, line 119: Rename to something like `useAccessiblePortal`

dkilgore-eightfold (changes requested):

dkilgore-eightfold (approved):

PR #417, opened 2022-10-12
fix(dropdown): modify flip behavior for filterable Select

atapshalkar-eightfold:
## SUMMARY:
Fixing flip behavior for filterable Select. 
When dropdown is opened upwards, the filtered dropdown would condense upwards instead of downwards. Example:
image

With this fix, the filtered dropdown will condense in the correct direction. 
image
image

NA

https://eightfoldai.atlassian.net/browse/ENG-31392
https://eightfoldai.atlassian.net/browse/ENG-31991 (clone)

## CHANGE TYPE:

-   [x] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
WIP

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -58,6 +58,7 @@ export const Dropdown: FC<DropdownProps> = React.memo(
         onClickOutside,
         width,
         height,
+        inputRef,
     }) => {
         const [mergedVisible, setVisible] = useMergedState<boolean>(false, {
             value: visible,
@@ -71,7 +72,7 @@ export const Dropdown: FC<DropdownProps> = React.memo(
             useFloating({
                 placement,
                 strategy: positionStrategy,
-                middleware: [flip(), fOffset(offset), shift()],
+                middleware: [fOffset(offset), flip(), shift()],
             });
 
         const toggle: Function =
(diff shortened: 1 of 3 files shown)


src/components/Dropdown/Dropdown.types.ts
```diff
@@ -90,4 +90,8 @@ export interface DropdownProps {
      * Manually control the width of the dropdown
      */
     width?: number;
+    /**
+     * Include dropdown textbox ref if using Select component
+     */
+    inputRef?: MutableRefObject<HTMLInputElement>;
```
dkilgore-eightfold, line 96: nit: alphabetize.

dkilgore-eightfold (approved):

src/components/Dropdown/Dropdown.tsx
```diff
@@ -71,7 +72,7 @@ export const Dropdown: FC<DropdownProps> = React.memo(
             useFloating({
                 placement,
                 strategy: positionStrategy,
-                middleware: [flip(), fOffset(offset), shift()],
+                middleware: [fOffset(offset), flip(), shift()],
```
dkilgore-eightfold, line 75: One interaction bug I found, and it's minor, is the direction of the animation isn't flipped. If this change could also address it that would be good, but not a blocker as the current bug is more impactful to the UX. We may address this in a future change if needed.

https://user-images.githubusercontent.com/99700808/195420954-14dfd30d-55d3-4722-bb52-a653dd1da2c1.mp4

dkilgore-eightfold: Perhaps this might do the trick --> https://floating-ui.com/docs/flip#flipalignment

atapshalkar-eightfold: Let's address this in another ticket/PR so as to not block on this. Let's also keep `fOffset` as the first argument to follow convention as mentioned in the docs.

src/components/Dropdown/Dropdown.types.ts
atapshalkar-eightfold: resolved

PR #427, opened 2022-10-27
fix: search box: pressing enter key clears text when clearable is true

dkilgore-eightfold:
## SUMMARY:
Handle clear button keyboard event so it doesn't trigger by default. It now requires focus.

Also updates buttons to newer system ui button with transparency. 

ENG-32974

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:

- Pull the PR branch and run `yarn` and `yarn storybook`. Verify the SearchBox and TextInput stories behave as expected.
- Open the ConfigProvider story and tab around to activate keyboard styles
- Go to the SearchBox story and click on the SearchBox and type some text. Pressing enter should trigger a bunko search, not clear the text. Tabbing to the clear icon and pressing enter should clear the text.


src/components/Inputs/input.module.scss
```diff
@@ -1469,13 +1465,6 @@
 
         .expandable-thumb {
             position: absolute;
-            left: -4px;
```
mfazil-eightfold, line 1472: I hope you've tested this. Earlier, the position was required because of the expandable transition effect.

dkilgore-eightfold: Yes, tested and it's good with the smaller size.

jhoward-eightfold (approved):

PR #430, opened 2022-11-03
feat: pagination: add ability to hide last page

ychhabra-eightfold:
## SUMMARY:
feat: pagination: add ability to hide last page

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/components/Pagination/Pager.tsx
+++ b/src/components/Pagination/Pager.tsx
@@ -25,6 +25,7 @@ export const Pager: FC<PagerProps> = React.forwardRef(
             quickNextIconButtonAriaLabel,
             quickPreviousIconButtonAriaLabel,
             simplified = false,
+            showLast = true,
             ...rest
         },
         ref: Ref<HTMLUListElement>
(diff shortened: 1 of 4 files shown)


src/components/Pagination/Pager.tsx
```diff
@@ -238,7 +239,7 @@ export const Pager: FC<PagerProps> = React.forwardRef(
                             />
                         </li>
                     )}
-                {pageCount > 1 && (
+                {pageCount > 1 && showLast && (
```
dkilgore-eightfold, line 242: Is the purpose of this that jumping this many pages does not work well with the specific DB being used? Or is it more about the last page being undefined in that DB? If it's the latter, I recommend tracking this in the array of pages and excluding the last. Say we get 100 pages - if showLast is false, this button would be page 99.

dkilgore-eightfold, line 242: Is the purpose of this that jumping this many pages does not work well with the specific DB being used? Or is it more about the last page being undefined in that DB? If it's the latter, I recommend tracking this in the array of pages and excluding the last. Say we get 100 pages - if showLast is false, this button would be page 99.

dkilgore-eightfold (approved):

PR #452, opened 2022-12-08
fix: input border color

ktorlakonda-eightfold:
## SUMMARY:
- fixing octuple styling based on yair's comments on a11y report
https://docs.google.com/spreadsheets/d/1TJhjfguoKzFpOZLeBXuVhz-EAclYSz6v/edit#gid=57133884&fvid=1123169165

https://eightfoldai.atlassian.net/browse/ENG-34711

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:


src/styles/themes/_default-theme.scss
```diff
@@ -165,7 +165,7 @@
     --grey-color-100: #05070a;
     --grey-color: #343c4c;
     --background-color: #fff;
-    --border-color: #bec2ca;
+    --border-color: #69717f;
```
jhoward-eightfold, line 168: Is it really the case that we want to change all usage of var(--border-color)? There are 25 instances of this variable in octuple codebase

this has far reaching impact that goes beyond the scope of pagination buttons. This probably needs to be done with more precision. See my comment on https://github.com/EightfoldAI/vscode/pull/33155

dkilgore-eightfold: @ktorlakonda-eightfold Yair was referring specifically to the Input component border color and isn't relevant to any other component. This line needs to be reverted.

dkilgore-eightfold: Please move this change to a local variable in `src/components/Inputs/input.module.scs`s and ensure it only effects the default border color for `rectangle`, `pill` and `underline` variants and you should be good.

dkilgore-eightfold (changes requested):

src/styles/themes/_default-theme.scss
ktorlakonda-eightfold: updated to use local variable
cc: @dkilgore-eightfold

src/components/Inputs/input.module.scss
```diff
@@ -11,6 +11,7 @@
         medium: $input-medium-size,
         large: $input-large-size,
     );
+    --border-color: #69717f;
```
jhoward-eightfold, line 14: you can have a localized var if you'd like, but name it something input specific and put it above with the others on lines above. Also, no need to set a specific value here, as you can assign a var from the existing palettes. Recommend this:

--input-border-color: var(--grey-color-60);

ktorlakonda-eightfold: done updated

src/components/Inputs/input.module.scss
```diff
@@ -1,7 +1,8 @@
 .input-wrapper {
+    --input-border-color: var(--grey-color-60);
```
dkilgore-eightfold, line 2: Now that the default is `--grey-color-60`, you'll need to create a second local variable, perhaps name it `--input-border-color-active` it should be `--grey-color-80`. Update all the active and hover border states to use this so we don't lose those states. These should be around the following lines:

339,
347,
1042,
1173,
1188

dkilgore-eightfold (changes requested):

src/components/Inputs/input.module.scss
ktorlakonda-eightfold: @dkilgore-eightfold fixed

dkilgore-eightfold (approved):

PR #472, opened 2022-12-19
fix: only generate a unique id once

ychhabra-eightfold:
## SUMMARY:
This fixes a new unique Id being generated on every render

## CHANGE TYPE:

- [ ] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Before: 

https://user-images.githubusercontent.com/95337653/208393928-9ebc0fa3-6a86-409e-8c66-c7ad8ae7fe8f.mov

After:

https://user-images.githubusercontent.com/95337653/208394228-66dd376d-efd6-4a07-9c61-f9afd8318d89.mov

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -1,7 +1,6 @@
 import React, {
   cloneElement,
   FC,
-  Ref,
   SyntheticEvent,
   useEffect,
   useImperativeHandle,
@@ -70,7 +69,7 @@ export const Dropdown: FC<DropdownProps> = React.memo(
       });
 
       const [closing, setClosing] = useState<boolean>(false);
-      const [dropdownId] = useState<string>(uniqueId('dropdown-'));
+      const dropdownId: string = uniqueId('dropdown-');
 
       let timeout: ReturnType<typeof setTimeout>;
       const { x, y, reference, floating, strategy, update, refs } = useFloating(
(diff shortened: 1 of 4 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -70,7 +69,7 @@ export const Dropdown: FC<DropdownProps> = React.memo(
       });
 
       const [closing, setClosing] = useState<boolean>(false);
-      const [dropdownId] = useState<string>(uniqueId('dropdown-'));
+      const dropdownId: string = uniqueId('dropdown-');
```
jhoward-eightfold, line 72: i thought useState was supposed to maintain value after getting initialized unless the setter gets called. Was this not to my understanding?

src/shared/utilities/uniqueId.ts
```diff
@@ -1,8 +1,10 @@
+import { useMemo } from 'react';
 /**
  * Generates a simple unique id
  * @param prefix The sting prefix
  */
 export const uniqueId = ((): ((prefix: string) => string) => {
   let counter: number = 0;
-  return (prefix: string): string => `${prefix}${++counter}`;
+  return (prefix: string): string =>
+    useMemo(() => `${prefix}${++counter}`, [prefix]);
```
jhoward-eightfold, line 9: won't this result in all inputs in the app lifecycle having the same "unique" id?

jhoward-eightfold (approved):
questions answered in Slack

PR #475, opened 2022-12-20
feat: second icon on button

mfazil-eightfold:
## SUMMARY:

Added two icons button

With text
image

When the text is empty
image

As Dropdown

https://user-images.githubusercontent.com/109717425/208925219-4856e4d2-4379-4b11-b64a-3ec161b9fd51.mov

Figma design:
https://www.figma.com/file/SlKRC7oKF7XZyHMv2op4ch/Octuple-DS-(Theme-2)?node-id=292%3A708&t=1XlyIUf3tu00a8ws-1
image

https://eightfoldai.atlassian.net/browse/ENG-38311

## CHANGE TYPE:

- [ ] Bugfix Pull Request
- [x] Feature Pull Request

## TEST COVERAGE:


src/components/Button/BaseButton.tsx
```diff
@@ -197,10 +199,14 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
     const getButtonLoader = (): JSX.Element =>
       loading && <Loader classNames={styles.loader} size={getLoaderSize()} />;
 
-    const getButtonIcon = (): JSX.Element => (
+    const getButtonIcon = (position = 'left'): JSX.Element => (
```
ychhabra-eightfold, line 202: can we create a different method for getting the second icon, will be cleaner

ychhabra-eightfold (changes requested):

src/components/Button/BaseButton.tsx
mfazil-eightfold: Will do

src/components/Button/button.module.scss
```diff
@@ -18,14 +18,26 @@
...
+    }
   }
 
   .icon + .button-text-small:not(:empty) {
     margin-left: $button-spacer-small;
+
+    + .icon-two {
+      margin-left: $button-spacer-small;
+    }
   }
```
jhoward-eightfold, line 41: need to account for RTL scenario also, in which case I presume the icons will switch sides?

mfazil-eightfold: RTL is handled by the button styles itself. 
image

src/components/Button/BaseButton.tsx
```diff
@@ -40,6 +40,7 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
       floatingButtonProps,
       htmlType,
       iconProps,
+      iconTwoProps,
```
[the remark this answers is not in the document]
mfazil-eightfold: I wanted to keep its consistency with TwoStateButton's `iconTwoProps`.  
https://github.com/EightfoldAI/octuple/blob/2b2c5710fe8160604ef78fa3bc1a1f3752e5e255/src/components/Button/TwoStateButton/TwoStateButton.tsx#L39

src/components/Button/BaseButton.tsx
```diff
@@ -10,7 +10,7 @@ import {
   InternalButtonProps,
   SplitButton,
 } from './';
-import { Icon, IconSize } from '../Icon';
+import { Icon, IconProps, IconSize } from '../Icon';
```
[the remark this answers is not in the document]
mfazil-eightfold: This is not directly related to the dropdown. 
Dropdown uses a button / an element and upon clicking, dropdown menu is shown. Hence, updating the dropdown file is not required.

src/components/Button/BaseButton.tsx
```diff
@@ -205,6 +207,18 @@ export const BaseButton: FC<InternalButtonProps> = React.forwardRef(
       />
     );
 
+    const getTwoButtonIcon = (): JSX.Element => (
```
ychhabra-eightfold, line 210:
```suggestion
    const getButtonIconTwo = (): JSX.Element => (
```

ychhabra-eightfold (changes requested):

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -150,6 +150,28 @@ const Dropdown_Button_Story: ComponentStory<typeof Dropdown> = (args) => {
...
 export const Dropdown_Button = Dropdown_Button_Story.bind({});
 
+const Dropdown_IconButton_Story: ComponentStory<typeof Dropdown> = (args) => {
+  const [visible, setVisibility] = useState(false);
+  return (
+    <Dropdown
+      {...args}
+      onVisibleChange={(isVisible) => setVisibility(isVisible)}
+    >
+      <DefaultButton
```
dkilgore-eightfold, line 160: Wouldn't it be more efficient to create another enum and prop called something like `variant` used by `TwoStateButton`? where the variants could be `filter` (current design), `default`, `primary`, `secondary`, `neutral`, and `system ui`. This may extend the current component in a logical way, more closely aligning it with other buttons beyond the filter scenario. It also has toggle behavior built in and may be more appropriate for use by `Dropdown`. The `TwoStateButton` already supports two icons.

dkilgore-eightfold (changes requested):

src/components/Dropdown/Dropdown.stories.tsx
mfazil-eightfold: We can add variants to TwoStateButton but that will take some bandwidth and PR review would again take some time.
Currently, we need this to support an upcoming release.

dkilgore-eightfold (approved):

PR #483, opened 2023-01-03
feat: stack: use flex gap instead of margins

ychhabra-eightfold:
## SUMMARY:
feat: stack: use flex gap instead of margins

## CHANGE TYPE:

- [ ] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/components/Stack/Stack.stories.tsx
+++ b/src/components/Stack/Stack.stories.tsx
@@ -86,7 +86,7 @@ export default {
       options: ['wrap', 'nowrap', 'wrap-reverse'],
       control: { type: 'radio' },
     },
-    gap: {
+    flexGap: {
       options: ['xxxs', 'xxs', 'xs', 's', 'm', 'l', 'xl', 'xxl', 'xxl', 'xxxl'],
       control: { type: 'select' },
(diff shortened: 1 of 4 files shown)


src/components/Stack/Stack.stories.tsx
```diff
@@ -86,7 +86,7 @@ export default {
       options: ['wrap', 'nowrap', 'wrap-reverse'],
       control: { type: 'radio' },
     },
-    gap: {
+    flexGap: {
```
dkilgore-eightfold, line 89: This would be a breaking change requiring a major release. Would it be better to leave the prop as is and change the functionality only?

dkilgore-eightfold (changes requested):

src/components/Stack/Stack.stories.tsx
dkilgore-eightfold, line 89: Need to update snaps

dkilgore-eightfold (changes requested):

dkilgore-eightfold (approved):

PR #491, opened 2023-01-04
fix: watch extra item width, padding and line to show for useMaxVisibleSections hook

ychhabra-eightfold:
## SUMMARY:
fix: watch extra item width, padding and line to show for useMaxVisibleSections hook

## CHANGE TYPE:

- [ ] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/hooks/useMaxVisibleSections.ts
+++ b/src/hooks/useMaxVisibleSections.ts
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 export const useMaxVisibleSections = (
   containerRef: React.MutableRefObject<HTMLElement>,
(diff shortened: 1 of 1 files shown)


src/hooks/useMaxVisibleSections.ts
```diff
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
```
mfazil-eightfold, line 1: why `React` import is needed?

ychhabra-eightfold: React import should be there on every react component

mfazil-eightfold: The hook does not require React import, only needed if you use any JSX.

ychhabra-eightfold: line no 4,5 have React.MutableRefObject, so it needs the react import

mfazil-eightfold (approved):

PR #508, opened 2023-01-19
chore: table: normalize aria label property name and add locs

dkilgore-eightfold:
## SUMMARY:
Adds remaining locs
Also some ui polish to prevent two pixel width button artifact of absolute layout

## CHANGE TYPE:
Chore

## TEST COVERAGE:

- [X] Tests for this change already exist

## TEST PLAN:
Pull the PR branch and run `yarn` and `yarn storybook`. Verify the `Table` stories behave as expected.

### DIFF
--- a/src/components/Table/Internal/Body/Scroller.tsx
+++ b/src/components/Table/Internal/Body/Scroller.tsx
@@ -24,8 +24,8 @@ export const Scroller = React.forwardRef(
       scrollBodyRef,
       stickyOffsets,
       scrollHeaderRef,
-      scrollLeftAriaLabel,
-      scrollRightAriaLabel,
+      scrollLeftAriaLabelText,
+      scrollRightAriaLabelText,
     }: ScrollerProps<RecordType>,
     ref: ForwardedRef<ScrollerRef>
   ) => {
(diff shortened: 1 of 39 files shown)


src/components/Table/Internal/Body/Scroller.tsx
```diff
@@ -194,7 +194,7 @@ export const Scroller = React.forwardRef(
             path: IconName.mdiChevronLeft,
           }}
           onClick={() => onClick('left')}
-          ariaLabel={scrollLeftAriaLabel}
+          ariaLabel={scrollLeftAriaLabelText}
```
jhoward-eightfold, line 197: these maybe work, but its come to my attention recently that aria attributes in react shouldn't be camelcased
https://reactjs.org/docs/accessibility.html
>Note that all aria-* HTML attributes are fully supported in JSX. Whereas most DOM properties and attributes in React are camelCased, these attributes should be hyphen-cased (also known as kebab-case, lisp-case, etc) as they are in plain HTML:

jhoward-eightfold (approved):

src/components/Table/Internal/Body/Scroller.tsx
dkilgore-eightfold: They work :) Might not in React 19 or 20, so we've some time for cleanup.

jhoward-eightfold: Now that I'm looking at this again, these are props which then get passed into the actual aria-label implementation. we're good here i bet

PR #517, opened 2023-01-27
feat: optionally show empty dropdown

fmehta-eightfold:
## SUMMARY:
In somecases empty dropdown is not required in the UI

ENG-39498

## CHANGE TYPE:

- [ ] Bugfix Pull Request
- [x] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Passing `showEmptyDropdown = false` will not show the dropdown when `options.length === 0`
**Before**
image
**After**
image


src/components/Select/Select.tsx
```diff
@@ -535,7 +536,9 @@ export const Select: FC<SelectProps> = React.forwardRef(
             onVisibleChange={(isVisible) => setDropdownVisibility(isVisible)}
             overlay={isLoading ? spinner : <OptionMenu options={options} />}
             showDropdown={showDropdown}
-            visible={dropdownVisible}
+            visible={
+              dropdownVisible && (showEmptyDropdown || options.length > 0)
```
dkilgore-eightfold, line 540: Does this account for the type-ahead scenario and still show the 'No match found' result? Never showing makes sense for on click with no input, but should provide some information with a partially input text.

slee-eightfold:
needs_sandbox

slee-eightfold:
@fmehta-eightfold How do I test the changes on sandbox? I need to run it on PCS.

src/components/Select/Select.tsx
fmehta-eightfold: You are right, this is done to handle the case of `click with no input`, and the `No match found` message will be shown with partially input text.

fmehta-eightfold:
> @fmehta-eightfold How do I test the changes on sandbox? I need to run it on PCS.

Refer vscode/pull/35681 on how to test this.

slee-eightfold (approved):
Tested in PCS Apply Form with the following conditions and is working as expected:
- empty options & showEmptyDropdown == true
Screenshot 2023-02-01 at 10 02 50 PM

- empty options & showEmptyDropdown == false
Screenshot 2023-02-01 at 10 03 11 PM

- non-empty options && showEmptyDropdown == false
Screenshot 2023-02-01 at 10 03 46 PM

PR #531, opened 2023-02-02
feat: link: adds variants with var theme styles and improves api

dkilgore-eightfold:
## SUMMARY:
- Adds `neutral`, `secondary` and `disruptive` variants
- Improves styles and extends varTheme CSS vars
- Adds `underline` prop
- Adds `disabled` prop
- Adds `onClick` prop and handler for when `disabled`
- Improves styles and makes it so `Link` variants don't need an external width by updating the default display to `inline-block`
- Adds aria attributes
- Adds `Link` unit tests

https://user-images.githubusercontent.com/99700808/216424831-58ee874d-93c7-4e5a-8a2d-5619647357b3.mp4

ENG-39138

## CHANGE TYPE:

- [X] Bugfix Pull Request
- [X] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull the PR branch and run `yarn` and `yarn storybook`. Verify the `Link` stories behave as expected.


ypike-eightfold (approved):

src/styles/themes/_default-theme.scss
```diff
@@ -208,7 +208,25 @@
   --backdrop-filter: blur(5px);
 
   --anchor-color: var(--primary-color-70);
+  --anchor-color-disruptive: var(--disruptive-color-70);
```
jhoward-eightfold, line 211: feel like you should invert the naming here so that sets of colors are grouped up

anchor-neutral-color
anchor-neutral-hover-color
anchor-neutral-visited-color

src/components/Link/link.module.scss
```diff
@@ -1,15 +1,95 @@
 .link-style {
   cursor: pointer;
+  display: inline-block;
   font-family: $octuple-font-family;
   font-size: $text-font-size-2;
+  text-decoration: underline;
+  width: fit-content;
+
+  &.neutral {
```
jhoward-eightfold, line 9: feels like an scss loop waiting to happen

src/components/Link/link.module.scss
```diff
@@ -1,15 +1,95 @@
 .link-style {
   cursor: pointer;
+  display: inline-block;
   font-family: $octuple-font-family;
   font-size: $text-font-size-2;
+  text-decoration: underline;
```
jhoward-eightfold, line 6: based on a11y related chat with Yair, seems like we want to give anchors a slightly heavier font weight, and only underline on hover/focus

jhoward-eightfold (approved):

src/components/Link/link.module.scss
dkilgore-eightfold: Will add semibold style, I used an css approach to underline, so the second bit of feedback is what's implemented. However, the default is preserved as underlined in those cases where engineers just need a plain-old link.

src/components/Link/link.module.scss
dkilgore-eightfold: This is a SASS sibling selector syntax, scoped to `.link-style`, a loop is unlikely.

src/components/Link/link.module.scss
dkilgore-eightfold: discussed with @ypike-eightfold - updated to semibold.

PR #538, opened 2023-02-08
fix: do not provide default colors for ConfigProvider

jhoward-eightfold:
## SUMMARY:
ConfigProvider is currently providing default values unnecessarily. This makes it necessary to provide empty objects in config in order to avoid the undesired default values.
Screenshot 2023-02-08 at 2 43 11 PM

In this example, I'm trying to emulate a starbucks green theme, and if I don't provide an empty navbarTheme and tabsTheme then the defaults get copied in and I end up with blue tabs despite my varTheme.

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull down the PR and run storybook locally. Select a new primary and/or accent color in ConfigProvider story. With this change, the tabs get updated appropriately. Current undesirable behavior is that they stay blue based on the default values.

### DIFF
--- a/src/components/ConfigProvider/Theming/Theming.types.ts
+++ b/src/components/ConfigProvider/Theming/Theming.types.ts
@@ -66,12 +66,6 @@ export interface ThemeOptions {
    */
   name?: ThemeName;
   /**
-   * Use system theme or not
-   * @default false
-   * @experimental
-   */
-  useSystemTheme?: boolean;
-  /**
    * Define a custom theme palette
    * @type {OcBaseTheme}
    * @default null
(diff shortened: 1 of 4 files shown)


src/components/ConfigProvider/Theming/Theming.types.ts
```diff
@@ -65,12 +65,6 @@ export interface ThemeOptions {
    * @default blue
    */
   name?: ThemeName;
-  /**
-   * Use system theme or not
-   * @default false
-   * @experimental
-   */
-  useSystemTheme?: boolean;
```
jhoward-eightfold, line 73: this experimental option wasn't being used anywhere so I cleaned it up.

dkilgore-eightfold (approved):

PR #540, opened 2023-02-13
feat: add secondary button in menu item

ychhabra-eightfold:
## SUMMARY:
feat: add secondary button in menu item

## CHANGE TYPE:

- [ ] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
+++ b/src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
@@ -1,8 +1,9 @@
-import React, { FC } from 'react';
+import React, { FC, useRef } from 'react';
 import { MenuItemButtonProps } from '../MenuItem.types';
 import { MenuSize, MenuVariant } from '../../Menu.types';
 import { mergeClasses } from '../../../../shared/utilities';
 import { Icon } from '../../../Icon';
+import { ButtonShape, ButtonSize, NeutralButton } from '../../../Button';
 
 import styles from '../menuItem.module.scss';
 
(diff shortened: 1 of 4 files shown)


src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
```diff
@@ -46,7 +49,11 @@ export const MenuItemButton: FC<MenuItemButtonProps> = ({
 
   return (
     <button
```
dkilgore-eightfold, line 51: This is an invalid pattern, Suggest using a conditional wrapper or separate elements. Wrapping a `<button>` in a `<button>`   is not accessible.

dkilgore-eightfold (changes requested):

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
```diff
@@ -44,7 +46,35 @@ export const MenuItemButton: FC<MenuItemButtonProps> = ({
     },
   ]);
 
-  return (
+  return secondaryButtonProps ? (
+    <li role="menuitem" tabIndex={tabIndex} className={menuItemClasses}>
+      <span className={styles.menuSecondaryWrapper}>
+        <button
```
dkilgore-eightfold, line 52: Was tabbing around and noticed there are limited styles for this raw button, and the keyboard styles are not matching the rest of the system. At the time this component was created, the `SystemUIButton` didn't exist. This one should prob be replaced in a future change with `SystemUIButton` with a slight modification to its text color.

dkilgore-eightfold (approved):

dkilgore-eightfold:
@ypike-eightfold This change adds a feature not yet available in the DS Figma:

secondaryMenuButton

It adds an additional button within Menu for actions like delete.

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
ychhabra-eightfold: will move to SystemUIButton in the future

PR #541, opened 2023-02-14
fix: link: isolate css display inline block to variants other than default

dkilgore-eightfold:
## SUMMARY:
`display: inline-block` was causing layout issues in vscode app Navbar implementation. This change isolates the display type to explicitly defined variants only, fixing the focus-visible borders in the app for already existing primary button variants. Also removes the need for link variants to require hard-coded widths.

### DIFF
--- a/src/components/Link/link.module.scss
+++ b/src/components/Link/link.module.scss
@@ -1,6 +1,5 @@
 .link-style {
   cursor: pointer;
-  display: inline-block;
   font-family: $octuple-font-family;
   font-weight: $text-font-weight-semibold;
   font-size: $text-font-size-2;
@@ -78,6 +77,7 @@
   &.primary,
   &.secondary,
   &.disruptive {
+    display: inline-block;
     text-decoration: none;
 
     &:active:not(.disabled),


src/components/Link/link.module.scss
```diff
@@ -78,6 +77,7 @@
   &.primary,
   &.secondary,
   &.disruptive {
+    display: inline-block;
```
jhoward-eightfold, line 80: seems undesirable to have these variants provide a different display property compared to a standard link.

dkilgore-eightfold: FYI, removing this will break `focus-visible` styles for `Link`. It's not a regression as current versions `focus-visible` styles are busted unless `Link` is given either a `display` type or `width` during implementation, either as inline styles or a CSS override, which should not be necessary IMHO.

jhoward-eightfold (approved):

src/components/Link/link.module.scss
jhoward-eightfold: i dont' want the app having to manage it. We will proceed with this since the default (non-variant) version shouldn't really be used. perhaps we should make primary the default as a breaking change in the near future?

PR #546, opened 2023-02-22
fix: stepper: Handle decimal scrollLeft value for scroll buttons

prathi-eightfold:
## SUMMARY:
ScrollLeft property can have decimal values for systems using display scaling ([ref](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft)). Handle decimal values when deciding to disable scroll buttons.

https://eightfoldai.atlassian.net/browse/ENG-43824

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:

Before

https://user-images.githubusercontent.com/77000634/220738620-ed678fff-aaa5-45bb-9ff0-0d719a26c503.mov

After

https://user-images.githubusercontent.com/77000634/220738699-22f1fe25-318b-4551-be4b-f092e4069084.mov

### DIFF
--- a/src/components/Stepper/Stepper.tsx
+++ b/src/components/Stepper/Stepper.tsx
@@ -219,12 +219,16 @@ export const Stepper: FC<StepperProps> = React.forwardRef(
         setNextDisabled(
           htmlDir === 'rtl'
             ? steps?.scrollLeft === steps?.offsetWidth - steps?.scrollWidth
-            : steps?.scrollLeft === steps?.scrollWidth - steps?.offsetWidth
+            : Math.abs(
+                steps?.scrollWidth - steps?.offsetWidth - steps?.scrollLeft
+              ) < 1
         );
         setPreviousDisabled(steps?.scrollLeft === 0);
       } else {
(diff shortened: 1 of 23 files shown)


src/components/Stepper/Stepper.tsx
```diff
@@ -219,12 +219,16 @@ export const Stepper: FC<StepperProps> = React.forwardRef(
         setNextDisabled(
           htmlDir === 'rtl'
             ? steps?.scrollLeft === steps?.offsetWidth - steps?.scrollWidth
```
dkilgore-eightfold, line 221: Math.abs should also work for rtl.

dkilgore-eightfold (changes requested):

dkilgore-eightfold (approved):

dkilgore-eightfold:
Would be extra if you add a CSS modifier to give the small variant paddles some top margin. If not now, I can do this in a future change.

PR #547, opened 2023-02-22
fix: nudge: removes unnecessary border calculations in css

dkilgore-eightfold:
## SUMMARY:
Using a CSS variable inside of a nested `calc` function isn't valid SASS, this change removes these property overrides to fix the `Size` nudge animation, because border is appended to the client rect, these overrides are unnecessary.

https://user-images.githubusercontent.com/99700808/220772475-8745647d-5bb4-4e11-80e0-0a5c44deb2df.mp4

ENG-44290

## CHANGE TYPE:

- [X] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:
N/A CSS only

## TEST PLAN:
Pull the PR branch and run `yarn` and `yarn storybook`. Verify the `PrimaryButton` story behaves as expected when given the following `nudgeProps`:

```
{
  "animation": "size",
  "delay": 5000,
  "enabled": true,
  "iterations": 5
}
```

### DIFF
--- a/src/components/Button/button.module.scss
+++ b/src/components/Button/button.module.scss
@@ -506,9 +506,6 @@
         border-width: var(--button-primary-border-width);
         border-style: var(--button-primary-border-style);
         border-color: var(--button-primary-border-color);
-        height: calc(100% + calc(var(--button-primary-border-width) * 2));
-        margin: calc(var(--button-primary-border-width) * -1);
-        width: calc(100% + calc(var(--button-primary-border-width) * 2));
       }
     }
   }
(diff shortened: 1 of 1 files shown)


src/components/Button/button.module.scss
```diff
@@ -506,9 +506,6 @@
         border-width: var(--button-primary-border-width);
         border-style: var(--button-primary-border-style);
         border-color: var(--button-primary-border-color);
-        height: calc(100% + calc(var(--button-primary-border-width) * 2));
-        margin: calc(var(--button-primary-border-width) * -1);
-        width: calc(100% + calc(var(--button-primary-border-width) * 2));
```
jhoward-eightfold, line 511: i could see why sass wouldn't like this.

jhoward-eightfold (approved):

PR #553, opened 2023-02-24
fix: popup: do not toggle when no content or disabled

dkilgore-eightfold:
## SUMMARY:
Initial Popup change omitted this condition, also ensures image flicker does not happen in Popup story


jhoward-eightfold (approved):

ypike-eightfold (commented):
Thanks.

PR #560, opened 2023-03-07
Avatar status item text and tooltip support

fmehta-eightfold:
## SUMMARY:
- Add `Avatar` status item **text** and **tooltip** support (this is needed because we required `Avatar+Tooltip` support in `AvatarGroup` but it was not working because of use of `cloneElement` in `AvatarGroup`)
- Improve `Avatar` status item placement CSS
- Fix undesired `Avatar` status item displacement in `AvatarGroup` due to border
- Fix `Avatar` outline when `src` is passed and `type === 'round'`

ENG-40399

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [x] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull the PR branch and run yarn and yarn storybook. Verify the Avatar Group stories behave as expected.

Avatar status item text
image

Avatar tooltip
image

Avatar status item displacement in AvatarGroup - **before** and **after** the fix
image
image

`Avatar` outline when `src` is passed and `type === 'round'` - **before** and **after** the fix
image
image

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -167,6 +167,7 @@ const AvatarStatusItems: FC<BaseAvatarProps> = React.forwardRef(
                     fontSize: statusItemProps.size,
                     color: statusItemProps.color,
                   }}
+                  className={styles.avatarStatusItemText}
                 >
                   {statusItemProps.text}
                 </span>
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -14,6 +14,8 @@ import {
 import { mergeClasses } from '../../shared/utilities';
 import { Icon } from '../Icon';
 import { useCanvasDirection } from '../../hooks/useCanvasDirection';
+import { Tooltip } from '../Tooltip';
+import { ConditionalWrapper } from '../../shared/utilities';
 
 export const AVATAR_THEME_SET = [
   styles.red,
@@ -263,6 +265,7 @@ export const Avatar: FC<AvatarProps> = React.forwardRef(
       hashingFunction,
       theme,
       randomiseTheme,
+      tooltipProps = undefined,
     },
     ref: Ref<HTMLDivElement>
   ) => {
@@ -301,20 +304,30 @@ export const Avatar: FC<AvatarProps> = React.forwardRef(
 
     if (src) {
       return (
-        <div ref={ref} style={wrapperContainerStyle} className={classNames}>
-          <img
-            src={src}
-            className={imageClasses}
-            alt={alt}
-            width={size}
-            height={size}
-          />
-          <AvatarStatusItems
-            outline={calculatedOutline}
-            size={size}
-            statusItems={statusItems}
-          />
-        </div>
+        <ConditionalWrapper
+          condition={tooltipProps !== undefined}
+          wrapper={(
+            children: React.ReactElement<
+              any,
+              string | React.JSXElementConstructor<any>
+            >
+          ): JSX.Element => <Tooltip {...tooltipProps}>{children}</Tooltip>}
+        >
+          <div ref={ref} style={wrapperContainerStyle} className={classNames}>
+            <img
+              src={src}
+              className={imageClasses}
+              alt={alt}
+              width={size}
+              height={size}
+            />
+            <AvatarStatusItems
+              outline={calculatedOutline}
+              size={size}
+              statusItems={statusItems}
+            />
+          </div>
+        </ConditionalWrapper>
       );
     }
 
@@ -322,38 +335,58 @@ export const Avatar: FC<AvatarProps> = React.forwardRef(
 
     if (iconProps) {
       return (
-        <AvatarIcon
-          iconProps={iconProps}
+        <ConditionalWrapper
+          condition={tooltipProps !== undefined}
+          wrapper={(
+            children: React.ReactElement<
+              any,
+              string | React.JSXElementConstructor<any>
+            >
+          ): JSX.Element => <Tooltip {...tooltipProps}>{children}</Tooltip>}
+        >
+          <AvatarIcon
+            iconProps={iconProps}
+            classNames={wrapperClasses}
+            style={wrapperContainerStyle}
+            fontSize={fontSize}
+            ref={ref}
+          >
+            <AvatarStatusItems
+              outline={calculatedOutline}
+              size={size}
+              statusItems={statusItems}
+            />
+          </AvatarIcon>
+        </ConditionalWrapper>
+      );
+    }
+
+    return (
+      <ConditionalWrapper
+        condition={tooltipProps !== undefined}
+        wrapper={(
+          children: React.ReactElement<
+            any,
+            string | React.JSXElementConstructor<any>
+          >
+        ): JSX.Element => <Tooltip {...tooltipProps}>{children}</Tooltip>}
+      >
+        <AvatarFallback
           classNames={wrapperClasses}
           style={wrapperContainerStyle}
-          fontSize={fontSize}
           ref={ref}
+          hashingFunction={hashingFunction}
+          theme={theme}
+          randomiseTheme={randomiseTheme}
         >
+          {children}
           <AvatarStatusItems
             outline={calculatedOutline}
             size={size}
             statusItems={statusItems}
           />
-        </AvatarIcon>
-      );
-    }
-
-    return (
-      <AvatarFallback
-        classNames={wrapperClasses}
-        style={wrapperContainerStyle}
-        ref={ref}
-        hashingFunction={hashingFunction}
-        theme={theme}
-        randomiseTheme={randomiseTheme}
-      >
-        {children}
-        <AvatarStatusItems
-          outline={calculatedOutline}
-          size={size}
-          statusItems={statusItems}
-        />
-      </AvatarFallback>
+        </AvatarFallback>
+      </ConditionalWrapper>
     );
   }
 );
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -162,7 +162,18 @@ const AvatarStatusItems: FC<BaseAvatarProps> = React.forwardRef(
                 ? { 'aria-label': statusItemProps.ariaLabel }
                 : {})}
             >
-              {statusItemProps.text ?? ''}
+              {statusItemProps.text ? (
+                <span
+                  style={{
+                    fontSize: `calc(2px + ${statusItemProps.size})`,
+                    color: statusItemProps.color,
+                  }}
+                >
+                  {statusItemProps.text}
+                </span>
+              ) : (
+                ''
+              )}
               <Icon {...statusItemProps} />
             </div>
           );
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -1,4 +1,4 @@
-import React, { FC, Ref, useMemo } from 'react';
+import React, { FC, Ref, useEffect, useMemo, useRef, useState } from 'react';
 
 // Styles:
 import styles from './avatar.module.scss';
@@ -11,11 +11,10 @@ import {
   StatusItemsPosition,
   StatusItemsProps,
 } from './';
-import { mergeClasses } from '../../shared/utilities';
+import { ConditionalWrapper, mergeClasses } from '../../shared/utilities';
 import { Icon } from '../Icon';
 import { useCanvasDirection } from '../../hooks/useCanvasDirection';
 import { Tooltip } from '../Tooltip';
-import { ConditionalWrapper } from '../../shared/utilities';
 
 export const AVATAR_THEME_SET = [
   styles.red,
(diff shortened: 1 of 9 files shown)


src/components/Avatar/Avatar.tsx
```diff
@@ -70,55 +72,43 @@ const AvatarStatusItems: FC<BaseAvatarProps> = React.forwardRef(
...
-      itemPos: StatusItemsPosition,
-      itemProps: StatusItemsProps,
-      wrapperPadding: string | number
+      itemPos: StatusItemsPosition
     ): React.CSSProperties => {
       const outlineWidth: string = outline?.outlineWidth ?? '0px'; // Avatar outline width
-      const avatarWidth: string = size;
-      const itemWidth: string = `(${itemProps.size} + (2 * ${wrapperPadding}))`; // Status item width
+      const outlineOffset: string = outline?.outlineOffset ?? '0px'; // Avatar outline offset
+      const radius: string = `calc((${size} + ${outlineWidth} + ${outlineOffset}) / 2)`;
```
dkilgore-eightfold, line 79: `AvatarGroup` items require outlines too, in your screen shot the outline was removed - will this change also work when group items have outlines to indicate status?

src/components/Avatar/Avatar.tsx
```diff
@@ -175,6 +162,18 @@ const AvatarStatusItems: FC<BaseAvatarProps> = React.forwardRef(
...
+                    fontSize: `calc(2px + ${statusItemProps.size})`,
+                    color: statusItemProps.color,
+                  }}
+                >
+                  {statusItemProps.text}
+                </span>
+              ) : (
+                ''
+              )}
               <Icon {...statusItemProps} />
```
dkilgore-eightfold, line 177: The `margin` between the `text` and the `Icon` is not sufficient. They look too close together.

src/components/Avatar/Avatar.tsx
```diff
@@ -135,16 +125,13 @@ const AvatarStatusItems: FC<BaseAvatarProps> = React.forwardRef(
               key={position}
               ref={ref}
               style={{
```
dkilgore-eightfold, line 127: Should these be moved into the css module?

src/components/Avatar/Avatar.tsx
```diff
@@ -175,6 +162,18 @@ const AvatarStatusItems: FC<BaseAvatarProps> = React.forwardRef(
...
                 : {})}
             >
+              {statusItemProps.text ? (
+                <span
+                  style={{
+                    fontSize: `calc(2px + ${statusItemProps.size})`,
+                    color: statusItemProps.color,
+                  }}
+                >
+                  {statusItemProps.text}
```
dkilgore-eightfold, line 172: This may be problematic if the user has `40,000` unread messages, etc. and won't scale. As long as it's compressed to `40k` we should be good. Though your implementation here should ensure a max width to avoid an edge case where a dev allows a large number.

src/components/Avatar/Avatar.tsx
```diff
@@ -175,6 +162,18 @@ const AvatarStatusItems: FC<BaseAvatarProps> = React.forwardRef(
                 ? { 'aria-label': statusItemProps.ariaLabel }
                 : {})}
             >
+              {statusItemProps.text ? (
+                <span
+                  style={{
+                    fontSize: `calc(2px + ${statusItemProps.size})`,
```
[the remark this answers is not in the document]
dkilgore-eightfold: On that note, would it be better to use css vars and put this in the module, referencing a class name?

src/components/Avatar/Styles/group.scss
```diff
@@ -7,7 +7,10 @@
 
   &.spaced {
     :not(:first-child):not(.avatar-group-tooltip) {
-      margin-left: -$space-xxxs;
+      margin-left: $space-xxxs;
+    }
+    & .avatar {
```
dkilgore-eightfold, line 12: `AvatarGroup` needs to support border too. This should be removed.

dkilgore-eightfold: Pass the border thickness from SASS to TypeScript an example may be found in `Slider` of how this is done.

dkilgore-eightfold (changes requested):

src/components/Avatar/Avatar.tsx
```diff
@@ -135,16 +125,13 @@ const AvatarStatusItems: FC<BaseAvatarProps> = React.forwardRef(
               key={position}
               ref={ref}
               style={{
+                display: 'flex',
                 position: 'absolute',
-                borderRadius: '50%',
+                borderRadius: '999px',
                 background:
                   statusItemProps.backgroundColor ?? 'var(--white-color)',
```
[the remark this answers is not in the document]
dkilgore-eightfold: These inline styles should be moved to SASS module. As this is a UI library, any use of CSS over inline style is preferred as a best practice.

src/components/Avatar/Styles/group.scss
dkilgore-eightfold: From there you may be able to account for the offset.

src/components/Avatar/Avatar.tsx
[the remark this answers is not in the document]
dkilgore-eightfold: Usage of scoped global CSS variables is encouraged along with the existing avatar theme in `_default-theme.scss`. Making these changes in the future can be complex and increase risk of regressions. Best to do it now.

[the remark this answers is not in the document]
dkilgore-eightfold: https://github.com/EightfoldAI/octuple/blob/main/src/styles/themes/_default-theme.scss#L619-L623

src/components/Avatar/Styles/group.scss
dkilgore-eightfold: Exported from SCSS:
https://github.com/EightfoldAI/octuple/blob/main/src/components/Slider/slider.module.scss#L30-L43

Typed:
https://github.com/EightfoldAI/octuple/blob/main/src/components/Slider/Slider.types.ts#L6-L32

Referenced:
https://github.com/EightfoldAI/octuple/blob/main/src/components/Slider/Slider.tsx#L220

src/components/Avatar/Avatar.tsx
dkilgore-eightfold: See comments below.

src/components/Avatar/Avatar.tsx
```diff
@@ -135,16 +125,13 @@ const AvatarStatusItems: FC<BaseAvatarProps> = React.forwardRef(
               key={position}
               ref={ref}
               style={{
+                display: 'flex',
                 position: 'absolute',
-                borderRadius: '50%',
+                borderRadius: '999px',
```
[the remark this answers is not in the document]
fmehta-eightfold: It does not help

`50%` and `100%`
image

`999px`
image

src/components/Avatar/Avatar.tsx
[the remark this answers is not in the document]
fmehta-eightfold: Right, using `statusItemProps.size` seems to be more appropriate.

src/components/Avatar/Avatar.tsx
fmehta-eightfold: No margin was added because the design also does not have it

src/components/Avatar/Avatar.tsx
```diff
@@ -301,59 +301,93 @@ export const Avatar: FC<AvatarProps> = React.forwardRef(
...
-            outline={calculatedOutline}
-            size={size}
-            statusItems={statusItems}
-          />
-        </div>
+        <ConditionalWrapper
+          condition={tooltipProps !== undefined}
+          wrapper={(
+            children: React.ReactElement<
+              any,
```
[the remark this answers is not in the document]
fmehta-eightfold: I had referred [AvatarGroup.tsx#L85](https://github.com/EightfoldAI/octuple/blob/main/src/components/Avatar/AvatarGroup.tsx#L85)

src/components/Avatar/Avatar.tsx
fmehta-eightfold: image

Enforcing max-width will result in visual loss of information.
Consider the three cases in the above image:
1. All information is visible. But, the width is more. User can report in case it is a blocker
2. Only partial information is visible. User is very likely to report it. Two solutions possible to overcome this issue:
    1. use something like `200k` - not possible from within the component as we do not know what text the dev will want to have in it
    2. show a tooltip - not suggested as it will complicate the UI
3. All information is visible and the width is also decent enough.

If we enforce max-width then case 2 will always occur and be an issue for the end user.
Occurrence of case 1 and 3 depends on the developer.

So, I think we should not restrict the width.

src/components/Avatar/Avatar.tsx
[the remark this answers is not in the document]
dkilgore-eightfold: kk, move into module via a class.

src/components/Avatar/Avatar.tsx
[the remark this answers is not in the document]
dkilgore-eightfold: You can pass size into a classname, look at this example --> https://github.com/EightfoldAI/octuple/blob/main/src/components/Stepper/Stepper.tsx#L289

src/components/Avatar/Avatar.tsx
dkilgore-eightfold: Correct, but kerning is a thing.

src/components/Avatar/Avatar.tsx
dkilgore-eightfold: What about a small Avatar? This will not scale.

dkilgore-eightfold: In the case of a large number, it would be best to just not display the number via a typeScript helper method that checks the num length.

src/components/Avatar/Avatar.tsx
dkilgore-eightfold: Maybe like 2px of space? It'll make a big difference.

src/components/Avatar/Avatar.tsx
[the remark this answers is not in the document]
fmehta-eightfold: Done
For future reference - https://stackoverflow.com/a/18795153

src/components/Avatar/Avatar.tsx
dkilgore-eightfold: @ypike-eightfold Please provide some guidance in Figma.

src/components/Avatar/Avatar.tsx
fmehta-eightfold: Done

src/components/Avatar/Avatar.tsx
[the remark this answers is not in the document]
fmehta-eightfold: Updated `React.ReactElement<...>` to `React.ReactNode`

src/components/Avatar/Avatar.tsx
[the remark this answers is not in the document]
fmehta-eightfold: Updated

src/components/Avatar/Avatar.tsx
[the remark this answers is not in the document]
fmehta-eightfold: Done

src/components/Avatar/Avatar.tsx
fmehta-eightfold: Yes, it will work

src/components/Avatar/Avatar.tsx
fmehta-eightfold: Done

src/components/Avatar/Styles/group.scss
fmehta-eightfold: Reverted this change because the status item shifting was fixed by other two changes:
1. use of `transform` in placement of status items
2. added `:not(.avatar-status-item)` to `src/components/Avatar/Styles/group.scss`

src/components/Avatar/Avatar.tsx
```diff
@@ -175,6 +169,19 @@ const AvatarStatusItems: FC<BaseAvatarProps> = React.forwardRef(
                 ? { 'aria-label': statusItemProps.ariaLabel }
                 : {})}
             >
+              {showStatusItemText ? (
+                <span
```
dkilgore-eightfold, line 173: You may use `getComputedStyle` of the span element by ref, something like:
```
const styles = getComputedStyle(ref.current);
console.log(styles.fontSize);
```

Create a ref for that span, get the computed style's `fontSize` and you should be good to go. Prob only need to do that in a `useMemo` as we care when the avatar is rendered.

In the memoized block, once you have that fontSize string, you may then `parseInt(n, 10)`, to convert that value to a number for the compare. Use the compare to update the `useState` boolean to either display or not display the count text via `showStatusItemText`.

dkilgore-eightfold: That way you may check if the font size is below 12px, adhere to the design system and hide the text when it's too small to be read.

dkilgore-eightfold (approved):

src/components/Avatar/Avatar.tsx
dkilgore-eightfold: Please try and make these changes prior to merging, everything else LGTM.

fmehta-eightfold: Done

PR #604, opened 2023-04-24
fix(avatar group): avatar group count, maxProps fixes

karurst-eightfold:
## SUMMARY:
fix(avatar group): avatar group count, maxProps fixes.
The maxProps count is not being properly considered to show the correct count.
Having wrong counts shown when sent as children or avatarListProps -> fixed.
Avatar Group completely null rendered or moot when the numChildren available less then maxProps set -> fixed.

https://eightfoldai.atlassian.net/browse/ENG-50821

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
1.) Check the below stories of Avatar Group, the earlier wrongly shown counts are properly shown and in case of the octuple stories of Exceed MaxCount than children all avatars must render inline.

Screenshot 2023-04-25 at 1 01 40 AM

### DIFF
--- a/src/components/Avatar/AvatarGroup.stories.tsx
+++ b/src/components/Avatar/AvatarGroup.stories.tsx
@@ -72,7 +72,6 @@ const sampleList: User[] = [
 
 const Basic_Story: ComponentStory<typeof AvatarGroup> = (args) => (
   <AvatarGroup
-    {...args}
     animateOnHover
     maxProps={{
       count: 4,
@@ -86,6 +85,7 @@ const Basic_Story: ComponentStory<typeof AvatarGroup> = (args) => (
         theme: TooltipTheme.dark,
       },
     }}
+    {...args}
   >
     <Avatar
       alt={imageProps.alt}
@@ -202,9 +202,10 @@ export const Basic = Basic_Story.bind({});
 
 export const Basic_Spaced = Basic_Story.bind({});
 
+export const Basic_Max_Props_Exceed_Children = Basic_Story.bind({});
+
 const List_Story: ComponentStory<typeof AvatarGroup> = (args) => (
   <AvatarGroup
-    {...args}
     animateOnHover
     avatarListProps={{
       items: sampleList,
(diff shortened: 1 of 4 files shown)


src/components/Avatar/AvatarGroup.tsx
```diff
@@ -111,21 +109,25 @@ export const AvatarGroup: React.FC<AvatarGroupProps> = React.forwardRef(
       </ConditionalWrapper>
     );
 
-    if (avatarListProps && maxCount && maxCount < numChildren) {
+    if (avatarListProps) {
+      const inlineCount = maxCount
```
dkilgore-eightfold, line 113: Please apply the same fixes to the `if (!avatarListProps)` condition too. That way the maxCount won't be required to render if the group only contains a single item.

dkilgore-eightfold (changes requested):

src/components/Avatar/AvatarGroup.tsx
karurst-eightfold: have made the recommended changes dylan.

src/components/Avatar/AvatarGroup.stories.tsx
```diff
@@ -202,9 +202,10 @@ export const Basic = Basic_Story.bind({});
 
 export const Basic_Spaced = Basic_Story.bind({});
 
+export const Basic_Max_Props_Exceed_Children = Basic_Story.bind({});
+
 const List_Story: ComponentStory<typeof AvatarGroup> = (args) => (
```
dkilgore-eightfold, line 207: nit: Update `maxProps` `tooltipProps.content` to something that matches the UI, like `User 6, User 7` or simply 'This is a tooltip.`.

dkilgore-eightfold (approved):

src/components/Avatar/AvatarGroup.stories.tsx
karurst-eightfold: doing.

karurst-eightfold: done.

PR #606, opened 2023-04-26
feat: adding style option for Empty description

dcoblentz-eightfold:
## SUMMARY:
Add descriptionClassNames property on Empty component

https://eightfoldai.atlassian.net/browse/ENG-51296

## CHANGE TYPE:
https://eightfoldai.atlassian.net/browse/ENG-51296

- [ ] Bugfix Pull Request
- [X] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
I need to include additional styling on the Empty component description text. The test case I've added should be sufficient to verify that this behaves as expected.

### DIFF
--- a/src/components/Empty/Empty.tsx
+++ b/src/components/Empty/Empty.tsx
@@ -79,7 +79,10 @@ export const Empty: FC<EmptyProps> = React.forwardRef(
         {title && <div className={styles.emptyTitle}>{title}</div>}
         {description && (
           <div
-            className={`${styles.emptyDescription} ${descriptionClassNames}`}
+            className={mergeClasses([
+              styles.emptyDescription,
+              descriptionClassNames,
+            ])}
           >
             {description}
           </div>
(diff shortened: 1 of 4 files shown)


src/components/Empty/Empty.tsx
```diff
@@ -77,7 +78,11 @@ export const Empty: FC<EmptyProps> = React.forwardRef(
         </div>
         {title && <div className={styles.emptyTitle}>{title}</div>}
         {description && (
-          <div className={styles.emptyDescription}>{description}</div>
+          <div
+            className={`${styles.emptyDescription} ${descriptionClassNames}`}
```
dkilgore-eightfold, line 82: nit: Rather than a templated string, you could use mergeClasses.

dkilgore-eightfold (approved):

src/components/Empty/Empty.tsx
dcoblentz-eightfold: Done

PR #608, opened 2023-05-01
fix: dropdown: use id from clone element props if present

dkilgore-eightfold:
## SUMMARY:

- Fixes `Select` focus steal on init by checking `firstRender` in the options change `useEffect`
- Fixes bug in `Dropdown` where the inherited id of the locally cloned element was being overridden by the component
- Fixes bug in smartApply where re-render caused the `defaultValue` to be `''`
- Fixes bug in smartApply where the option `role` was being overridden by `menuitem` default

ENG-51660

## CHANGE TYPE:

- [X] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull the PR branch and run `yarn`, `yarn test` and `yarn storybook`. Verify all tests pass and the `Dropdown`, `Menu` and `Select` stories behave as expected.

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -74,8 +74,9 @@ export const Dropdown: FC<DropdownProps> = React.memo(
 
       const [closing, setClosing] = useState<boolean>(false);
       const dropdownId: string = uniqueId('dropdown-');
-      const dropdownReferenceId: React.MutableRefObject<string> =
-        useRef<string>(`${dropdownId}reference`);
+      const [dropdownReferenceId, setReferenceElementId] = useState<string>(
+        `${dropdownId}reference`
+      );
 
       let timeout: ReturnType<typeof setTimeout>;
       const { x, y, reference, floating, strategy, update, refs, context } =
@@ -111,9 +112,8 @@ export const Dropdown: FC<DropdownProps> = React.memo(
       useOnClickOutside(
         refs.floating,
         (e) => {
-          const referenceElement: HTMLElement = document.getElementById(
-            dropdownReferenceId?.current
-          );
+          const referenceElement: HTMLElement =
+            document.getElementById(dropdownReferenceId);
           if (closeOnOutsideClick && closeOnReferenceClick) {
             toggle(false)(e);
           }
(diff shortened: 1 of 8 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -239,11 +239,14 @@ export const Dropdown: FC<DropdownProps> = React.memo(
           { [child.props.className]: child.props.className },
           { [styles.disabled]: disabled },
         ]);
+        if (child.props.id && dropdownReferenceId !== child.props.id) {
+          setReferenceElementId(child.props.id);
```
dcoblentz-eightfold, line 243: This method to update the state is async, but in the code below, you are assuming that the state will have been updated. Is this still safe? I'm still getting used to the react lifecycle, but it seems like the below code _may_ have the update or may not (depending on how react queues the state update), but also that setting the state would also trigger another render, in which case it would pick up the correct value the second time?

src/components/Select/Select.tsx
```diff
@@ -180,6 +180,12 @@ export const Select: FC<SelectProps> = React.forwardRef(
     // Update options on change
     useEffect(() => {
       onOptionsChange?.(getSelectedOptionValues(), getSelectedOptions());
+
+      // Determine first render to help determine the Select interaction is intentional
+      if (firstRender.current) {
+        firstRender.current = false;
+        return;
+      }
       if (multiple) {
```
dcoblentz-eightfold, line 189: Honestly, this is still worrying me a bit. Our intent here is to have this element get focus specifically when a user sets a value on it, but not when a value is programmatically set right? This check prevents the focus from getting set on the first focus, but I'm wondering what other cases could trigger focus

What if we do an immediate render while making an async call to get a list of available options which we then set a default value for?

Other possibility would be a form where someone sets a value on one property, which then triggers a different value on a select.

Can we do something here to be more explicit about doing this only when we _know_ it's been initiated by a user?

src/components/Dropdown/Dropdown.tsx
dkilgore-eightfold: The string compare takes care of this and ensures the state is only ever set one time.

src/components/Select/Select.tsx
dkilgore-eightfold: What are your thoughts?

dkilgore-eightfold: Technically there is no way to explicitly determine a user has initiated the on change event of the input, as this may be done programmatically. The above code only ever triggers when the input onchange event occurs. I'll investigate a different approach.

src/components/Dropdown/Dropdown.tsx
dkilgore-eightfold: `getReference` method is not async and the component is memoized.

src/components/Select/Select.tsx
dkilgore-eightfold: Added an additional compare for the previous value of `dropdownVisible` to improve the heuristics of the code.

src/components/Dropdown/Dropdown.tsx
dkilgore-eightfold: Added a comment clarifying the reason behind the compare.

dcoblentz-eightfold (approved):
I tested these locally and everything looks good!

PR #626, opened 2023-05-24
fix: pill: ensure pill heights match latest design specification

dkilgore-eightfold:
## SUMMARY:
Pill component height is determined by the `line-height` of its content because it may wrap via `lineclamp` prop. This change adjusts the `line-height` so it's not a float, updates `padding` as necessary to match the following spec:

`PillSize.Large` = `44px`
`PillSize.Medium` = `30px`
`PillSize.Small` and `PillSize.XSmall` = `24px`

The medium size `Pill` spec also requires the `border-radius` to be `6px`, as this isn't snapped to the base 4 grid, it will not be referenced from SASS variable definitions. Some padding also needs to be local to the module as the base 4 `line-height` effects that spacing relative to the `font-size`.

https://github.com/EightfoldAI/octuple/assets/99700808/6147123d-4db7-4c8c-ac34-8fbcb1cba9b3

[Figma Specification (Eightfold internal only)](https://www.figma.com/file/SlKRC7oKF7XZyHMv2op4ch/Octuple-DS-(Theme-2)?type=design&node-id=560-606&t=f51rWYLNhhyNCWQR-0)

ENG-51328

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

- [x] Tests for this change already exist (CSS only UI snaps updated with this change)

## TEST PLAN:
Pull the PR branch and run `yarn` and `yarn storybook`. Verify the `Pill` stories behave as expected.

### DIFF
--- a/src/components/Pills/pills.module.scss
+++ b/src/components/Pills/pills.module.scss
@@ -3,8 +3,8 @@
   --label: var(--blue-color);
   --hover-bg: var(--blue-color-30);
   --hover-label: var(--text-primary-color);
-  padding: $space-xxs $space-s;
-  border-radius: $border-radius-m;
+  padding: 5px $space-xs;
+  border-radius: 6px;
   transition: all $motion-duration-extra-fast $motion-easing-easeinout 0s;
   background: var(--bg);
   color: var(--label);
@@ -13,17 +13,23 @@
   align-items: center;
 
   .label {
-    font-family: $octuple-font-family;
-    font-size: $text-font-size-5;
+    font-family: var(--font-stack-full);
     font-weight: $text-font-weight-semibold;
+
+    &.large {
+      font-size: $text-font-size-5;
+      line-height: $text-line-height-3;
+    }
 
     &.medium {
       font-size: $text-font-size-3;
+      line-height: $text-line-height-2;
     }
 
     &.small,
     &.xsmall {
       font-size: $text-font-size-2;
+      line-height: $text-line-height-1;
     }
 
     &.line-clamp {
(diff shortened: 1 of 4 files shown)


src/components/Pills/pills.module.scss
```diff
@@ -3,8 +3,8 @@
   --label: var(--blue-color);
   --hover-bg: var(--blue-color-30);
   --hover-label: var(--text-primary-color);
-  padding: $space-xxs $space-s;
-  border-radius: $border-radius-m;
+  padding: 5px $space-xs;
+  border-radius: 6px;
```
jhoward-eightfold, line 7: do we want to expose as a css var? probably not 🤷

src/components/Pills/pills.module.scss
```diff
@@ -120,7 +126,25 @@
     --bg: var(--white-color);
     --label: var(--grey-color-60);
     --hover-bg: var(--grey-color-10);
-    border: 1px solid var(--grey-color-60);
+
+    &:not(.xsmall) {
+      border: 1px solid var(--grey-color-60);
```
jhoward-eightfold, line 131: semantic css var?

jhoward-eightfold (approved):

src/components/Pills/pills.module.scss
dkilgore-eightfold: Once merged I can update the other PR.

src/components/Pills/pills.module.scss
dkilgore-eightfold: Not for padding, but maybe we _should_ expose the border radius.

Issue #633, opened 2023-06-05
loading indicator should color according to button it is implemented in.


jhoward-eightfold:
Currently, loading indicator is hardcoded to primary color, which doesn't work on primary buttons.

![Screenshot 2023-06-05 at 2 41 44 PM](https://github.com/EightfoldAI/octuple/assets/106610186/66373a1f-5406-4fbd-88d1-caa6746646cb)

PR #636, opened 2023-06-06
fix: button: ensure loading dots use text color and enable two state loader

dkilgore-eightfold:
## SUMMARY:
Add `dotClassNames` prop to Loader
Ensure dots use button variant text colors
Adds loading prop to `TwoStateButton`

https://github.com/EightfoldAI/octuple/assets/99700808/d72a120b-b550-4c55-af97-983e69fbb329

https://github.com/EightfoldAI/octuple/issues/633

ENG-55278

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

- [x] Tests for this change already exist (snap updated)

## TEST PLAN:
Pull PR branch and run `yarn` and `yarn storybook`. Verify `loading` state for all `Button` stories behaves as expected.

### DIFF
--- a/src/components/Button/TwoStateButton/TwoStateButton.tsx
+++ b/src/components/Button/TwoStateButton/TwoStateButton.tsx
@@ -13,6 +13,7 @@ import {
 import { Badge } from '../../Badge';
 import { Icon, IconSize } from '../../Icon';
 import { InnerNudge, NudgeAnimation, NudgeProps } from '../Nudge';
+import { Loader, LoaderSize } from '../../Loader';
 import { Breakpoints, useMatchMedia } from '../../../hooks/useMatchMedia';
 import { mergeClasses } from '../../../shared/utilities';
 import { useCanvasDirection } from '../../../hooks/useCanvasDirection';
@@ -41,6 +42,7 @@ export const TwoStateButton: FC<TwoStateButtonProps> = React.forwardRef(
       iconOneProps,
       iconTwoProps,
       id,
+      loading = false,
       nudgeProps: defaultNudgeProps,
       onClick,
       shape = ButtonShape.Pill,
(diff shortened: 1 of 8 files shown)


src/components/Loader/loader.module.scss
```diff
@@ -37,8 +37,7 @@
     }
 
     &-rtl {
-      margin-left: $space-xxs;
-      margin-right: unset;
+      margin: 0 $space-xxxs;
     }
```
jhoward-eightfold, line 41: seemingly this rule becomes unnecessary

src/components/Loader/loader.module.scss
```diff
@@ -5,7 +5,7 @@
     --translate: -2px;
     background: var(--primary-color);
     border-radius: 50%;
-    margin-right: $space-xxs;
+    margin: 0 $space-xxxs;
```
jhoward-eightfold, line 8: seems like this will improve centering, nice

src/components/Button/TwoStateButton/TwoStateButton.tsx
```diff
@@ -177,6 +180,36 @@ export const TwoStateButton: FC<TwoStateButtonProps> = React.forwardRef(
...
+        loaderSize = LoaderSize.Small;
+      }
+      return loaderSize;
+    };
+
+    const getButtonLoader = (): JSX.Element =>
+      loading && (
+        <Loader
+          classNames={styles.loader}
+          dotClassNames={styles.loaderDot}
```
jhoward-eightfold, line 208: best i can tell from poking around in dev tools, you might be able to avoid the overhead of explicitly setting the color for each implementation by setting `background: currentColor`, and then it will match the text color of the given button

src/components/Loader/loader.module.scss
dkilgore-eightfold: Aiming to keep the rtl reference in css as it may be needed in the future.

src/components/Button/TwoStateButton/TwoStateButton.tsx
dkilgore-eightfold: Unfortunately, the bg color of loader is the bg color of button. In this case it would not work.

dkilgore-eightfold: There are loader implementations outside of button in vscode, we should keep the explicit color definitions

jhoward-eightfold (approved):

Issue #665, opened 2023-07-11
popup has trouble with events on touch screens


jhoward-eightfold:
When using a touchscreen, popups/tooltips won't trigger if the ref element is in focus.

https://github.com/EightfoldAI/octuple/assets/106610186/c519f330-d217-4613-9809-b27667fb7ec7

dkilgore-eightfold:
Tracking with ENG-58436

PR #668, opened 2023-07-17
fix: tooltip: fixes tooltip toggle on mobile devices

dkilgore-eightfold:
## SUMMARY:
Adds `mobileDetector` utility to detect iOS or Android user agents is `isIOS()` and `isAndroid()`
Adds `mobileDetector` unit tests
Updates `Tooltip` to read user agent and determine if `hover` `trigger` is allowed
Fixes `Popup` hover trigger to not dismiss when hovering off the reference element on to the `Popup` itself

Mobile:

https://github.com/EightfoldAI/octuple/assets/99700808/c58f7361-a3bb-48a5-a32a-64415f0c7d3c

Desktop:

https://github.com/EightfoldAI/octuple/assets/99700808/21e681da-7b41-479a-aafb-f5251ffe714f

https://github.com/EightfoldAI/octuple/issues/665

ENG-58436

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull the PR branch and run `yarn` and `yarn storybook`. Verify the `Popup` and `Tooltip` stories behave as expected.


src/components/Tooltip/Tooltip.tsx
```diff
@@ -103,6 +105,15 @@ export const Tooltip: FC<TooltipProps> = React.memo(
         `${tooltipId?.current}-reference`
       );
 
+      const isMobile: boolean = isIOS() || isAndroid();
```
jhoward-eightfold, line 108: so is the fix explicit to these mobile OSes? Is it possible to apply this fix when a device is generically touch enabled? Microsoft Surface comes to mind.

dkilgore-eightfold: I'll add a Surface method too that takes this into account, detecting based on touch only is not reliable.

src/shared/utilities/mobileDetector.ts
```diff
@@ -0,0 +1,29 @@
...
+    return false;
+  }
+  return /android/i.test(window.navigator.userAgent);
+};
+
+/**
+ * Returns true if and only if the user is on a touch device and is currently using its touch API.
+ * Used to determine interaciton with hybrid touch devices like Microsoft Surface.
+ */
+export const isTouching = (): boolean => {
```
jhoward-eightfold, line 27: consider a name update here. it doesnt check is user is actively "touching" but rather if the device supports "touch"... maybe `isTouchDevice` or `isTouchSupported`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -104,6 +107,18 @@ export const Tooltip: FC<TooltipProps> = React.memo(
...
```
jhoward-eightfold, line 110: might be a reasonable exported method like isTouching. most scenarios will want the generic over the specific os I would think

src/components/Tooltip/Tooltip.tsx
```diff
@@ -104,6 +107,18 @@ export const Tooltip: FC<TooltipProps> = React.memo(
...
+      let mergedTrigger: 'click' | 'hover' | 'contextmenu';
+
+      // First use the most reliable way of detecting iOS or Android,
+      // then test for touch interaction as a somewhat less reliable
+      // way to detect other hybrid devices.
+      if (isMobile || (!isMobile && isTouching())) {
```
jhoward-eightfold, line 116: in theory this now will work in chrome emulation as well?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -189,7 +205,9 @@ export const Tooltip: FC<TooltipProps> = React.memo(
       );
 
       const handleReferenceClick = (event: React.MouseEvent): void => {
-        event.stopPropagation();
+        if (type !== TooltipType.Default) {
+          event.stopPropagation();
```
jhoward-eightfold, line 209: so, if You put the tooltip on a clickable element will it still trigger the click event on touch? I'm guessing yes based on your attached videos, but I'm not 100%

jhoward-eightfold (approved):

src/components/Tooltip/Tooltip.tsx
dkilgore-eightfold: When it comes to screen readers we'll need the specificity.

src/components/Tooltip/Tooltip.tsx
dkilgore-eightfold: Will double check a console log is triggered on the element click. The video only represents the visible state, not the event handling.

dkilgore-eightfold: Made some updates, we are good on devices. Tooltip and Popup reference element onClick behave as expected without this change. Reverting and also applying to `handleReferenceKeyDown`

src/components/Tooltip/Tooltip.tsx
dkilgore-eightfold: Verified.

Issue #666, opened 2023-07-13
Issue with select component if not passing any options


shazare-eightfold:
I noticed a bug, if I don't specify options in select component of octuple then the following error occurs and also the page becomes unresponsive.
![image](https://github.com/EightfoldAI/octuple/assets/133693194/4b51f28f-6a46-4630-8403-db016263de1a)

dkilgore-eightfold:
Tracking with ENG-58585

PR #669, opened 2023-07-17
fix: select: adds fall back and optional chaining to avoid null reference errors at runtime

dkilgore-eightfold:
## SUMMARY:

- Adds options fallback to empty array where needed and some optional chaining
- Updates `placeholder` to show when `options` are null in` multiple` mode
- Adds unit tests covering null or empty array scenarios
- Updates `options` to ensure `id` alignment.

https://github.com/EightfoldAI/octuple/assets/99700808/4e1e18c7-81da-4707-be4e-10a158edfa2c

https://github.com/EightfoldAI/octuple/issues/666

ENG-58585

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull the PR branch and run `yarn` and `yarn storybook`. Verify the `Select` stories behave as expected when its `options` are `null` or `[]`.

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -110,11 +110,11 @@ export const Select: FC<SelectProps> = React.forwardRef(
     );
     const [dropdownVisible, setDropdownVisibility] = useState<boolean>(false);
     const [options, setOptions] = useState<SelectOption[]>(
-      _options.map((option: SelectOption, index: number) => ({
+      _options?.map((option: SelectOption, index: number) => ({
         selected: false,
         hideOption: false,
-        id: option.text + '-' + index,
-        object: option.object,
+        id: option?.text + '-' + index,
+        object: option?.object,
         role: 'option',
         ...option,
       }))
@@ -147,12 +147,12 @@ export const Select: FC<SelectProps> = React.forwardRef(
 
     const getSelectedOptionValues = (): SelectOption['value'][] => {
       return options
-        .filter((option: SelectOption) => option.selected)
-        .map((option: SelectOption) => option.value);
+        ?.filter((option: SelectOption) => option?.selected)
+        .map((option: SelectOption) => option?.value);
     };
 
     const getSelectedOptions = (): SelectOption['value'][] => {
-      return options.filter((option: SelectOption) => option.selected);
+      return options?.filter((option: SelectOption) => option?.selected);
     };
 
     const { count, filled, width } = useMaxVisibleSections(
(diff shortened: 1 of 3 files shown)


src/components/Select/Select.tsx
```diff
@@ -110,11 +110,11 @@ export const Select: FC<SelectProps> = React.forwardRef(
     );
     const [dropdownVisible, setDropdownVisibility] = useState<boolean>(false);
     const [options, setOptions] = useState<SelectOption[]>(
-      _options.map((option: SelectOption, index: number) => ({
+      _options?.map((option: SelectOption, index: number) => ({
```
jhoward-eightfold, line 113: do you want to fallback to an empty array so code can rely on this being an array?
```suggestion
      (_options || []).map((option: SelectOption, index: number) => ({
```

src/components/Select/Select.tsx
```diff
@@ -147,12 +147,12 @@ export const Select: FC<SelectProps> = React.forwardRef(
 
     const getSelectedOptionValues = (): SelectOption['value'][] => {
       return options
-        .filter((option: SelectOption) => option.selected)
-        .map((option: SelectOption) => option.value);
+        ?.filter((option: SelectOption) => option?.selected)
```
jhoward-eightfold, line 150: by falling back to array above you probably don't need these optional chains from here down

src/components/Select/Select.tsx
```diff
@@ -161,20 +161,20 @@ export const Select: FC<SelectProps> = React.forwardRef(
...
     useEffect(() => {
-      const selected: SelectOption[] = options.filter(
-        (opt: SelectOption) => opt.selected
+      const selected: SelectOption[] = options?.filter(
+        (opt: SelectOption) => opt?.selected
       );
       setOptions(
-        _options.map((option: SelectOption, index: number) => ({
-          selected: !!selected.find((opt) => opt.value === option.value),
+        _options?.map((option: SelectOption, index: number) => ({
```
jhoward-eightfold, line 173:
```suggestion
        (_options || []).map((option: SelectOption, index: number) => ({
```

src/components/Select/Select.tsx
```diff
@@ -161,20 +161,20 @@ export const Select: FC<SelectProps> = React.forwardRef(
...
+          selected: !!selected?.find((opt) => opt?.value === option?.value),
           hideOption: false,
-          id: option.text + index,
-          object: option.object,
+          id: option?.text + '-' + index,
```
jhoward-eightfold, line 176: perhaps loop functions can check that the type of option is correct and filter it out if not... things might go pretty sidewise if we are trying to support objects of the wrong type

src/components/Select/Select.tsx
```diff
@@ -490,24 +490,24 @@ export const Select: FC<SelectProps> = React.forwardRef(
     };
 
     const getPills = (): JSX.Element => {
-      const selected: SelectOption[] = options.filter(
-        (opt: SelectOption) => opt.selected
+      const selected: SelectOption[] = options?.filter(
```
jhoward-eightfold, line 493: similarly, if you provide an empty fallback, you can safely assume selected is an array below
```suggestion
      const selected: SelectOption[] = (options || []).filter(
```

src/components/Select/Select.tsx
dkilgore-eightfold: This sounds risky, shouldn't JSON validation be deferred to the host app?

dkilgore-eightfold: A few risks:

- Unnecessary parsing
- Duplicate validation impacting perf
- Runtime error if validation is missing some edge case

jhoward-eightfold (approved):

PR #682, opened 2023-08-07
feat: card: update card size prop and remove bunko story

dkilgore-eightfold:
## SUMMARY:
- Removes Outdated Story
- Adds `TODO` comment to add predefined card variants and update the one that currently exists
- Adds a small size (small used to map to medium via the `CardSize` enum)
- Updates padding for each size (`large`, `medium`, and `small`) to spec
- Adds unit tests

https://github.com/EightfoldAI/octuple/assets/99700808/afbd8299-766a-406d-8903-16d8d0e10e24

ENG-61008

## CHANGE TYPE:

- [ ] Bugfix Pull Request
- [x] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull the PR branch and run `yarn` and `yarn storybook`. Verify the `Card` story behaves as expected.

### DIFF
--- a/src/components/Card/card.module.scss
+++ b/src/components/Card/card.module.scss
@@ -5,7 +5,8 @@
   display: block;
   font-family: var(--card-font-family);
   gap: $space-m;
-  padding: $space-ml;
+  height: 260px;
+  padding: $space-m;
   width: 360px;
   position: relative;
 
@@ -30,11 +31,17 @@
 
   &-large {
     height: 278px;
+    padding: $space-ml;
   }
 
-  &-medium,
+  &-medium {
+    height: 260px;
+    padding: $space-m;
+  }
+
   &-small {
     height: 260px;
+    padding: $space-s;
   }
 
   &.disabled {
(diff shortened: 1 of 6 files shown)


src/components/Card/card.module.scss
```diff
@@ -5,7 +5,8 @@
   display: block;
   font-family: var(--card-font-family);
   gap: $space-m;
-  padding: $space-ml;
+  height: 260px;
```
jhoward-eightfold, line 8: still of the opinion that predefined heights for cards at the library level is too prescriptive

src/components/Card/card.module.scss
```diff
@@ -30,11 +31,17 @@
...
+
+  &-medium {
+    height: 260px;
+    padding: $space-m;
   }
 
-  &-medium,
   &-small {
     height: 260px;
+    padding: $space-s;
```
jhoward-eightfold, line 44: and regarding prescriptive heights, There is so little variance in these, hard to justify calling the property "size" when mostly it just tweaks padding.m My 2cents

jhoward-eightfold (approved):

src/components/Card/card.module.scss
dkilgore-eightfold: Maybe we'll remove in v3, for now some UI is prob depending on this.

src/components/Card/card.module.scss
dkilgore-eightfold: Needed to play it safe, reducing the height of `small` as it was previously `medium` was a risk.

PR #706, opened 2023-09-06
feat: inputs: add readonly support and its styles

dkilgore-eightfold:
## SUMMARY:

- Updates `TextInput`, `TextArea`, `SearchBox` `readonly` and `readOnlyProps`, enabling complex custom readonly implementations.
- Updates Select `readonly` to behave as expected
- Adds UTs

https://github.com/EightfoldAI/octuple/assets/99700808/75995ec6-fa42-49e3-ba8f-82bcfa0c5f29

ENG-60991

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [x] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull the PR branch and run `yarn` and `yarn storybook`. Verify the `Inputs` and `Select` stories behave as expected.

### DIFF
--- a/src/components/Inputs/Input.types.ts
+++ b/src/components/Inputs/Input.types.ts
@@ -95,11 +95,6 @@ export interface ReadOnlyProps {
    */
   clearable?: boolean;
   /**
-   * Whether readonly is enabled.
-   * @default false
-   */
-  enabled?: boolean;
-  /**
    * The readonly icon props.
    */
   iconProps?: IconProps;
@@ -153,7 +148,11 @@ export interface TextAreaProps
    * text area readonly.
    * @default false
    */
-  readonly?: boolean | Omit<ReadOnlyProps, 'clearable'>;
+  readonly?: boolean;
+  /**
+   * text area readonly props.
+   */
+  readOnlyProps?: Omit<ReadOnlyProps, 'clearable'>;
   /**
    * The text area required attribute.
    * @default false
@@ -351,7 +350,11 @@ export interface InputProps<T>
    * input readonly.
    * @default false
    */
-  readonly?: boolean | ReadOnlyProps;
+  readonly?: boolean;
+  /**
+   * input readonly props.
+   */
+  readOnlyProps?: ReadOnlyProps;
   /**
    * Resets the input value cache.
    */
--- a/src/components/Inputs/Input.types.ts
+++ b/src/components/Inputs/Input.types.ts
@@ -134,6 +134,7 @@ export interface TextAreaProps
     | 'iconProps'
     | 'iconButtonProps'
     | 'alignIcon'
+    | 'readonly'
   > {
   /**
    * The text area is expandable.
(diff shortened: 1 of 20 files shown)


ypike-eightfold (approved):

src/components/Select/select.module.scss
```diff
@@ -93,14 +93,14 @@ $multi-select-count-offset: 54px;
     }
 
     .select-clear-button {
-      bottom: 0;
-      height: 44px;
+      bottom: $space-xxs;
+      height: 36px;
```
jhoward-eightfold, line 97: why did this get smaller?

src/components/Select/Select.types.ts
```diff
@@ -104,6 +104,11 @@ export interface SelectProps
    * The Select input custom class names.
    */
   inputClassNames?: string;
+  /**
+   * The Select input is readonly.
+   * @default false
+   */
+  inputReadOnly?: boolean;
```
jhoward-eightfold, line 111: does this require the word "input"? would love to dial down the verbosity if possible

src/components/Select/Select.tsx
```diff
@@ -261,6 +262,12 @@ export const Select: FC<SelectProps> = React.forwardRef(
       }
     }, [clear, dropdownVisible, filterable]);
 
+    useEffect(() => {
+      if (inputReadOnly && dropdownVisible) {
+        setDropdownVisibility(false);
```
jhoward-eightfold, line 267: nice edge case handling. is it worth a test?

src/components/Inputs/TextInput/TextInput.stories.tsx
```diff
@@ -164,6 +167,12 @@ Text_Input.args = {
   name: 'myTextInput',
   numbersOnly: false,
   placeholder: 'Placeholder text',
+  readonly: {
+    clearable: false,
+    enabled: false,
```
jhoward-eightfold, line 172: Is it worthwhile to separate the boolean from the additional props to avoid the type checking?

src/components/Inputs/TextArea/TextArea.tsx
```diff
@@ -217,6 +237,27 @@ export const TextArea: FC<TextAreaProps> = React.forwardRef(
       debouncedChange(e);
     };
 
+    const getIconSize = (): IconSize => {
+      let iconSize: IconSize;
+      if (largeScreenActive) {
```
jhoward-eightfold, line 242: would be nice to have a hook which simply returns current screen size ('lg', 'md', 'sm', 'xs' etc), and then a mapping could be used instead of if/else. I would guess this would be useful in other components?

src/components/Inputs/Input.types.ts
```diff
@@ -323,7 +351,7 @@ export interface InputProps<T>
    * input readonly.
    * @default false
    */
-  readonly?: boolean;
+  readonly?: boolean | ReadOnlyProps;
```
jhoward-eightfold, line 354: I did this review in reverse order, so this is called out below as well. Can we separate the props and the boolean? It would read nicer without the type checking, etc

src/components/Select/select.module.scss
dkilgore-eightfold: Need room for focus-visible rect.

src/components/Inputs/TextArea/TextArea.tsx
dkilgore-eightfold: We have such a hook, and this could be a mapping as the properties here are returned by a hook, but will consider this for a future change as it's code style only.

jhoward-eightfold (approved):

Issue #738, opened 2023-10-23
Update label wrapping and required mark on Form.Item


dcoblentz-eightfold:
Example:
<img width="1123" alt="image" src="https://github.com/EightfoldAI/octuple/assets/119535119/1fb8feac-8c57-4e36-b797-2661bc03a2c9">

PR #747, opened 2023-11-14
fix: form: fixes required and optional marks so they display inline with the label text

dkilgore-eightfold:
## SUMMARY:
- The `Form.Item` `required` and `requiredMark` were displaying as a separate columns, such that the required asterisk was centering on wrapped text. This change fixes this bug so that the asterisk displays inline with the text
- Fixes a bug where the required asterisk and colon could never be displayed together 
- Updates stories to use `<Button variant={foo} />` rather than `<FooButton />`

https://github.com/EightfoldAI/octuple/assets/99700808/733d62d4-8ab4-4bf8-b98e-e489591f8749

https://github.com/EightfoldAI/octuple/issues/738

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

- [x] Tests for this change already exist (HTML and CSS only reflected in Storybook - functionality remains unchanged)

## TEST PLAN:
Pull the PR branch and run `yarn` and `yarn storybook`. Verify the `Form` stories behave as expected with regards to wrapping text and `required` and `requiredMark="optional"` props.


src/components/Form/form.module.scss
```diff
@@ -125,27 +125,27 @@
...
-        .form-hide-required-mark {
+        &.form-item-required:not(.form-item-required-mark-optional),
+        &.form-item-required-mark-optional:not(.form-item-required) {
+          margin-left: $space-xxs;
+        }
+
+        &.form-item-required:not(.form-item-required-mark-optional):after {
+          display: inline-block;
+          margin-right: $space-xxs;
+          color: var(--text-primary-color);
```
dcoblentz-eightfold, line 140: we have a pretty common ask from customer for the required * to be colored red instead of text primary color. Would it be possible to set up something like:

`--form-item-required-mark-color: var(--text-primary-color);` (or whatever name makes the most sense to you)

And then we easily customize this color via theming?

dkilgore-eightfold: Will do.

dcoblentz-eightfold (approved):
Thanks for the changes!

PR #823, opened 2024-04-24
feat: accordion: deep accessibility updates

dkilgore-eightfold:
## SUMMARY:
- Removes multiple click target concern by moving overall clickable area inside the header, under the content via `zIndex`
- Updates `focus-visible` styles to work with the new layout
- Accounts for custom interactive content via selectors, such that it remains selectable or clickable
- Checks if custom content is a `string`, wraps it in a `span` so it remains selectable
- Removes a duplicate `{...expandButtonProps}` attribute
- Updates unit tests
- Updates stories

Lighthouse report:
accordionDeepA11yPassLighthouse

Axe report:
accordionDeepA11yPassAxe

ENG-87358

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [x] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Pull the PR branch and run `yarn` and `yarn storybook`. Verify the `Accordion` stories behave as expected.

### DIFF
--- a/src/components/Accordion/Accordion.tsx
+++ b/src/components/Accordion/Accordion.tsx
@@ -15,6 +15,7 @@ import ThemeContext, {
 } from '../ConfigProvider/ThemeContext';
 import {
   AccordionBodyProps,
+  AccordionLocale,
   AccordionProps,
   AccordionShape,
   AccordionSize,
@@ -24,6 +25,10 @@ import { Badge } from '../Badge';
 import { Button, ButtonShape, ButtonVariant } from '../Button';
 import { Icon, IconName } from '../Icon';
 import { eventKeys, mergeClasses, uniqueId } from '../../shared/utilities';
+import LocaleReceiver, {
+  useLocaleReceiver,
+} from '../LocaleProvider/LocaleReceiver';
+import enUS from './Locale/en_US';
 
 import styles from './accordion.module.scss';
 import themedComponentStyles from './accordion.theme.module.scss';
@@ -34,7 +39,9 @@ export const AccordionSummary: FC<AccordionSummaryProps> = ({
   badgeProps,
   children,
   classNames,
+  collapseAriaLabelText,
   disabled,
+  expandAriaLabelText,
   expandButtonProps,
   expanded,
   expandIconProps,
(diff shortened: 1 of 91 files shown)


ypike-eightfold (approved):

src/components/Accordion/Accordion.tsx
```diff
@@ -77,24 +84,28 @@ export const AccordionSummary: FC<AccordionSummaryProps> = ({
...
-      className={headerClassnames}
-      onClick={onClick}
-      onKeyDown={handleKeyDown}
-      id={`${id}-header`}
-      role="button"
-      tabIndex={0}
-      {...rest}
-    >
+    <div className={headerClassnames} id={`${id}-header`} {...rest}>
+      <div
```
jhoward-eightfold, line 88: I have mixed feelings about this. Now it is less clear what the content is that is being expanded. And the reading from VoiceOver is quite confusing to me:
"Collapse content, expanded, button"
"Expand content, collapsed, button"
should we externalize the aria-label here?

In this case, what would we want/expect the button to announce?
Screenshot 2024-04-24 at 2 58 45 PM

dkilgore-eightfold: The aria labels are externalized via props, but we need to ensure there is a default label regardless.

dkilgore-eightfold: As the Header content is a `React. ReactNode`, there's no guarantee it's a `string`. I can make an update applying the header content to the aria label if it's a `typeof` `string`, then fallback to the default aria labels. That should suffice for basic implementations. Ideally the aria labels will be provided at implementation time via props, but engineers may miss this one.

dkilgore-eightfold: For example, in the screen above, the content isn't a string. It contains 2 elements that in turn, contain strings, and from an aria-label perspective, would not be valid.

dkilgore-eightfold: Pushed updates:

Content design feedback:
Open content, collapsed, button
Close content, expanded, button

Also reversed the default/externalized aria labels such that they are more intuitively aligned with the state of the buttons:

`collapseAriaLabelText` = "Close content"
`expandAriaLabelText` = "Open content"

dkilgore-eightfold: The hard truth is, in order to be 100% accessible while retaining the same level of useful information, aria labels will need to be provided as an implementation task.

dkilgore-eightfold: Another option would be to only pass the component name as the default aria label:

Content design feedback:
Accordion, collapsed, button
Accordion, expanded, button

LMK

jhoward-eightfold (approved):

PR #825, opened 2024-04-29
fix: carousel and tooltip: preventTouchMoveDefault false for tooltips inside carousels 

gclark-eightfold:
## SUMMARY:
- Fixes an issue on mobile / touch devices where the tooltip `preventTouchMoveDefault` behavior would prevent carousels from being draggable.
- Fixes the JSDoc annotation for the tooltip `touchInteraction` prop to correctly reflect the default: `TooltipTouchInteraction.TapAndHold`

> NOTE: Once this is released in a new version, https://github.com/EightfoldAI/vscode/pull/58487 and https://github.com/EightfoldAI/vscode/pull/58660 can & should be reverted. 

N/A

https://eightfoldai.atlassian.net/browse/ENG-86212

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Load the Octuple storybook for Carousel on a mobile device, or on desktop using the mobile view inside of the Chrome Dev Tools (`⌘ ⇧ M`)
In the new "Scroller with Tooltips" story:
- [ ] Tapping on the text inside the card shows a tooltip with the same text.
- [ ] Pressing on the text and dragging left/right moves the carousel cards.

### DIFF
--- a/src/components/Carousel/Carousel.stories.tsx
+++ b/src/components/Carousel/Carousel.stories.tsx
@@ -11,6 +11,7 @@ import {
 import { Button, ButtonShape, ButtonSize, ButtonVariant } from '../Button';
 import { Card } from '../Card';
 import { IconName } from '../Icon';
+import { Tooltip } from '../Tooltip';
 import { useCanvasDirection } from '../../hooks/useCanvasDirection';
 
 export default {
@@ -310,7 +311,9 @@ Scroller.args = {
             width: '100%',
           }}
         >
-          {item.name}
+          <Tooltip content={item.name} portal>
+            {item.name}
+          </Tooltip>
         </div>
       </Card>
     )),
@@ -335,7 +338,9 @@ Scroller_Single.args = {
             width: '100%',
           }}
         >
-          {item.name}
+          <Tooltip content={item.name} portal>
+            {item.name}
+          </Tooltip>
         </div>
       </Card>
     )),
@@ -361,7 +366,9 @@ Scroller_Custom_Buttons.args = {
             width: '100%',
           }}
         >
-          {item.name}
+          <Tooltip content={item.name} portal>
+            {item.name}
+          </Tooltip>
         </div>
       </Card>
     )),
--- a/src/components/Carousel/Carousel.stories.tsx
+++ b/src/components/Carousel/Carousel.stories.tsx
@@ -11,7 +11,7 @@ import {
 import { Button, ButtonShape, ButtonSize, ButtonVariant } from '../Button';
 import { Card } from '../Card';
 import { IconName } from '../Icon';
-import { Tooltip } from '../Tooltip';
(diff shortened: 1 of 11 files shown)


src/components/Carousel/Carousel.stories.tsx
```diff
@@ -310,7 +311,9 @@ Scroller.args = {
             width: '100%',
           }}
         >
-          {item.name}
+          <Tooltip content={item.name} portal>
```
dkilgore-eightfold, line 314: Suggest copy/pasting the Tooltip with truncation here for a better E2E example of a typical Card text. Maybe via a quick concatenation of the item and index with some Lorem Ipsum - or the same text from the [Truncation Detection](https://github.com/EightfoldAI/octuple/blob/main/src/components/Tooltip/Tooltip.stories.tsx#L131) story.

src/components/ConfigProvider/ParentComponentsContext.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { createContext, FC, useContext, useMemo } from 'react';
```
dkilgore-eightfold, line 1: If this is to be a abstracted context that may be used by any parent, Move this into https://github.com/EightfoldAI/octuple/tree/main/src/components/ConfigProvider

src/components/Carousel/Carousel.tsx
```diff
@@ -829,87 +830,92 @@ export const Carousel: FC<CarouselProps> = React.forwardRef(
...
+                            return React.cloneElement(
+                              child as React.ReactElement<any>,
+                              {
+                                active: active === index ? true : false,
+                                direction: direction,
+                                key: index,
+                              }
+                            );
+                          }
+                          return null;
```
dkilgore-eightfold, line 891: A UT covering this line would be helpful.

dkilgore-eightfold (changes requested):

src/components/ConfigProvider/ParentComponentsContext.tsx
dkilgore-eightfold: Exporting this and its props may also be helpful.

src/components/Carousel/Carousel.stories.tsx
```diff
@@ -375,3 +383,53 @@ Scroller_Custom_Buttons.args = {
...
+                textAlign: 'center',
+                textOverflow: 'ellipsis',
+                WebkitBoxOrient: 'vertical',
+                WebkitLineClamp: 4,
+                whiteSpace: 'pre-line',
+                width: '100%',
+              }}
+            >
+              {TOOLTIP_INFO_TEXT}
+            </div>
```
dkilgore-eightfold, line 425: Replace with useTruncate hook implementation. Would be a good demo of the hook only.

dkilgore-eightfold (approved):

PR #829, opened 2024-05-10
feat: support tree shaking

gclark-eightfold:
## SUMMARY:
Adds tree-shaking support for apps using Octuple. Core changes:
1. Add explicit `sideEffects` to `package.json` for CJS tree shaking.
2. Switch from Webpack to Rollup to support exporting ESM. 

For the Webpack -> Rollup change, it _may_ be possible to export ESM from webpack, but most sources I found said exporting ESM from webpack was [not supported](https://github.com/webpack/webpack/issues/2933). Additionally, after examining quite a number of specifically react component libraries, it seemed like Rollup was the most common tool for this job.

> Testing on Eightfold for the `careerHub` entry point only yielded a `-20.9%` decrease in the bundle size for the octuple chunk compared to v2.51.1

Here is the bundle analysis diff in Blindfold.ai after following the steps outlined in the Test Plan
### Before
Before

### After
After

NA

ENG-55139

## CHANGE TYPE:

- [ ] Bugfix Pull Request
- [x] Feature Pull Request

## TEST COVERAGE:

1. Clone the https://github.com/EightfoldAI/blindfold repo locally
2. In the root of the repo, add the following `.env` file. (If you have one already, just add `ANALYZE_BUNDLES=1`
```
ANALYZE_BUNDLES=1
BACKEND_BASE_URL="https://app.eightfold.ai"
NEXT_PUBLIC_HOST_URL="https://localhost:8000"
NEXTAUTH_URL="https://localhost:8000"
```
2. Run `git checkout gclark/octuple-bundle-test` to pull [these changes](https://github.com/EightfoldAI/blindfold/pull/453/files).
3. Install deps with `pnpm install`. (Run `npm install -g pnpm` if you do not have pnpm installed)
4. Run `pnpm build --filter nextjs` to build the Blindfold app.
5. Once complete open the file located in `apps/nextjs/.next/analyze/client.html` in your web browser.
    - [x] The bundle analysis does not show the entire `octuple.js` bundle present.
6. Run `pnpm start --filter nextjs` and go to http://localhost:8000/octuple
    - [x] Verify the Octuple primary button is present on the page and styled correctly.
7. Remove the patch with `pnpm patch-remove @eightfold.ai/octuple`
8. Run `pnpm build --filter nextjs` again.
9. Once complete open the file located in `apps/nextjs/.next/analyze/client.html` in your web browser.
    - [x] The bundle analysis **does** show the entire `octuple.js` bundle present.


.github/workflows/build.yml
```diff
@@ -11,7 +11,7 @@ on:
       - main-next
 
 env:
-  NODE_VERSION: 16.14.2
+  NODE_VERSION: 20.13.0
```
dkilgore-eightfold, line 14: 👍🏼

package.json
```diff
@@ -83,6 +108,11 @@
     "@commitlint/config-conventional": "16.2.1",
     "@ctrl/tinycolor": "3.4.0",
     "@pmmmwh/react-refresh-webpack-plugin": "0.5.3",
+    "@rollup/plugin-commonjs": "25.0.7",
+    "@rollup/plugin-node-resolve": "15.2.3",
+    "@rollup/plugin-terser": "0.4.4",
+    "@rollup/plugin-typescript": "11.1.6",
+    "@rollup/plugin-url": "^8.0.2",
```
dkilgore-eightfold, line 115: remove tophat

package.json
```diff
@@ -180,6 +211,14 @@
...
     "resolve": "1.20.0",
     "resolve-url-loader": "4.0.0",
+    "rollup": "4.17.2",
+    "rollup-plugin-dts": "6.1.0",
+    "rollup-plugin-peer-deps-external": "2.2.4",
+    "rollup-plugin-postcss": "4.0.2",
+    "rollup-plugin-preserve-directives": "^0.4.0",
+    "rollup-plugin-progress": "^1.1.2",
+    "rollup-plugin-typescript-paths": "^1.5.0",
+    "rollup-plugin-visualizer": "^5.12.0",
```
dkilgore-eightfold, line 221: Same, no tophat

package.json
```diff
@@ -2,18 +2,43 @@
   "name": "@eightfold.ai/octuple",
   "version": "2.51.1",
   "license": "MIT",
-  "main": "lib/octuple.js",
+  "sideEffects": [
+    "**/*.css"
+  ],
+  "exports": {
```
dkilgore-eightfold, line 8: Lets add Individual components to this. That way we're set to scale for experimental and vNext components.

package.json
```diff
@@ -2,18 +2,43 @@
...
     "type": "git",
     "url": "https://github.com/EightfoldAI/octuple"
   },
   "scripts": {
     "storybook": "start-storybook -p 2022",
     "build-storybook": "rm -rf storybook-static && mkdir storybook-static && build-storybook -o storybook-static",
-    "build": "rm -rf lib && npm run lint && webpack --config webpack.prod.js --mode production",
-    "details": "webpack --config webpack.dev.js --mode development --stats-error-details",
-    "analyze": "webpack --config webpack.dev.js --mode production --json > stats.json",
+    "prebuild": "rm -rf lib && yarn lint",
```
dkilgore-eightfold, line 37: We could keep this in the build script, that way local build is a 1 liner

rollup.config.mjs
```diff
@@ -0,0 +1,94 @@
...
+        return;
+      }
+      warn(warning);
+    },
+  },
+  {
+    // TODO: audit if this is needed (or add the same for lib/locale.d.ts)
+    input: 'lib/octuple.d.ts',
+    output: [{ file: 'lib/octuple.d.ts', format: 'esm' }],
+    external: [/\.(css|less|scss)$/],
```
dkilgore-eightfold, line 91: No less here, prob can be removed.

rollup.config.mjs
```diff
@@ -0,0 +1,94 @@
...
+        dir: 'lib',
+        entryFileNames: '[name].mjs',
+        format: 'es',
+        sourcemap: false,
+      },
+    ],
+    plugins: [
+      progress(),
+      postcss({
+        modules: { localsConvention: 'camelCase' },
```
dkilgore-eightfold, line 38: Did you run the UTs locally, they may use webpack still, but should be verified.

dkilgore-eightfold (changes requested):

package.json
dkilgore-eightfold: Prob need to add like a public.ts in the same dir as index.ts each and only map to those exported from octuple.ts

dkilgore-eightfold:
Need to bump node in .codesanbox dir

dkilgore-eightfold:
@gclark-eightfold You may be correct:

https://github.com/webpack/webpack/issues/2933

rollup.config.mjs
gclark-eightfold: Tests seemed to all pass when running `yarn test`
Screenshot 2024-05-09 at 5 55 55 PM

package.json
gclark-eightfold: Don't think that will be necessary.

Do all the components we want to export individually follow a pattern like `src/components/**/index.{ts,tsx}`?
If so it's not super difficult to get these added

gclark-eightfold: After trying it out a bit, I think it's probably not necessary right now.

Additional exports can be added fairly trivially by adding:
```
preserveModules: true,
preserveModulesRoot: 'src',
```
to the output configurations in `rollup.config.mjs`, but unfortunately it doesn't seem to generate `index.mjs` and `index.cjs` files for each component, only the named files (e.g. `lib/components/Accordion/Accordion.cjs` exists, but `lib/components/Accordion/index.cjs` does not)

If that can be fixed without needing to go all out and add individual entry points for each export, then each entry point can be added like:
```json
{
    "exports": {
        "./Accordion": {
            "types": "./lib/components/Accordion/index.d.ts",
            "import": "./lib/components/Accordion/index.mjs",
            "require": "./lib/components/Accordion/index.cjs"
        }
    }
}
```

_This lengthy JSON could also be generated to avoid needing to create/update these by hand_

gclark-eightfold: Also, from looking around at other libraries (e.g. [redux](https://github.com/reduxjs/redux/blob/master/package.json), [radix-ui/primitives](https://github.com/radix-ui/primitives/blob/main/packages/react/slot/package.json)), it seems like most take the approach of either:
1. A monorepo with a `packages` directory, each exporting its own thing (components, react/svelte/<js lib> bindings, etc.)
2. A single package, but only exports the top-level `'.'` or limited additional entry points.

package.json
```diff
@@ -2,16 +2,40 @@
...
+  "sideEffects": [
+    "**/*.css"
+  ],
+  "exports": {
+    ".": {
+      "types": "./lib/octuple.d.ts",
+      "import": "./lib/octuple.mjs",
+      "require": "./lib/octuple.cjs"
+    },
+    "./lib/*": {
```
dkilgore-eightfold, line 14: Currently, the only other export is locale, so wouldn't "./lib/locale" suffice here? I don't recommend we export all from lib as it's not what we currently do.

package.json
```diff
@@ -282,8 +317,7 @@
     ]
   },
   "engines": {
-    "node": ">=16.13.0 < 17",
-    "npm": ">=8.5.0 < 9"
+    "node": ">=18.0.0"
```
dkilgore-eightfold, line 320: Modified this to >=16.0.0 to fix husky. Seems to work fine.

dkilgore-eightfold:
Also removed this line and its import to fix the circular dep issue accross bundles:

https://github.com/EightfoldAI/octuple/blob/main/src/shared/utilities/flexGapSupported.ts#L7

package.json
dkilgore-eightfold: @gclark-eightfold Exporting individual components as packages to NPM seems the way to go like [radix-ui/primitives](https://github.com/radix-ui/primitives/blob/main/packages/react/slot/package.json). However, the build script should be shared and updates will need to be made to the Actions to enable this. Does it need to be this complex? What we do with locale works with our current webpack config. Seems to me to be a limitation of rollup's module transpiler not considering Jest causing the 1 isssue we have found when running upstream UTs.

dkilgore-eightfold (approved):

Issue #866, opened 2024-08-07
Accordions provide two toggle buttons serving the same purpose


jhoward-eightfold:
This is confusing for screen reader users

PR #867, opened 2024-08-07
fix: stop screenreaders from reading duplicate accordion toggle btns

jhoward-eightfold:
## SUMMARY:
This change takes the toggle button out of the tabIndex and gives it a presentation role, relying on the full header click area to provide keyboard support and focus for the accordion toggle. it also points to the header content as a describedBy so its read in addition to the generic button label. Both of these can be overwritten on a case-by-case basis.

https://github.com/EightfoldAI/octuple/issues/866

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
- pull branch down locally and start storybook (`yarn storybook`)
- with screen reader on, use tab key to interact with accordion to observe improvements in behavior

### DIFF
--- a/src/components/Accordion/Accordion.tsx
+++ b/src/components/Accordion/Accordion.tsx
@@ -42,6 +42,7 @@ export const AccordionSummary: FC<AccordionSummaryProps> = ({
   collapseAriaLabelText,
   disabled,
   expandAriaLabelText,
+  expandButtonDescribedBy,
   expandButtonProps,
   expanded,
   expandIconProps,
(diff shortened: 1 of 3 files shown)


src/components/Accordion/Accordion.tsx
```diff
@@ -89,23 +90,27 @@ export const AccordionSummary: FC<AccordionSummaryProps> = ({
...
         className={styles.clickableArea}
         onClick={onClick}
         onKeyDown={handleKeyDown}
         role="button"
         tabIndex={0}
       ></div>
-      <div className={styles.accordionHeaderContainer}>
+      <div
+        id={`${id}-header-content`}
+        className={styles.accordionHeaderContainer}
```
sravichandran-eightfold, line 102: there is no aria-label here. What will aria-describedby do?

jhoward-eightfold: It will read the content of the div. In some cases that can be information overload. In simple cases its a nice clean readout. For the following it would read approximately "Accordion, Color themes, button, collapsed"

Screenshot 2024-08-07 at 7 01 48 AM

jhoward-eightfold: For complex cases I've provided the prop so you can unset it (in which case it will leverage the aria labe props for the button) or point it to some other (likely visuallyhidden) descriptor

sravichandran-eightfold (approved):

szhang-eightfold (approved):

PR #875, opened 2024-08-20
feat: feature flags: Add feature flag config for Octuple

dcoblentz-eightfold:
Add a feature flag config for octuple which can provide a mechanism to provide default behavior for Octuple components.

## SUMMARY:
This adds a feature flag config, and creates one flag that can control the way that Panels are rendered

## CHANGE TYPE:

- [ ] Bugfix Pull Request
- [X] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Open the panel in storybook, and toggle the `lazyLoadPanelContent` flag in the settings. You should be able to see that the panel contents are rendered or not rendered even when the panel is not visible

### DIFF
--- a/src/components/ConfigProvider/FeatureFlagProvider.tsx
+++ b/src/components/ConfigProvider/FeatureFlagProvider.tsx
@@ -3,7 +3,7 @@
 import React, { createContext, FC, useContext, useMemo } from 'react';
 
 export interface FeatureFlags {
-  lazyLoadPanelContent?: boolean;
+  panelLazyLoadContent?: boolean;
 }
 
 export interface FeatureFlagContextProps {
@@ -14,7 +14,7 @@ export interface FeatureFlagContextProps {
 const FeatureFlagContext = createContext<FeatureFlags>(undefined);
 
 const defaultFeatureFlagValues: FeatureFlags = {
-  lazyLoadPanelContent: false,
+  panelLazyLoadContent: false,
 };
 
 export const FeatureFlagContextProvider: FC<FeatureFlagContextProps> = ({
--- a/src/components/ConfigProvider/FeatureFlagProvider.tsx
+++ b/src/components/ConfigProvider/FeatureFlagProvider.tsx
@@ -1,0 +1,49 @@
+'use client';
+
+import React, { createContext, FC, useContext, useMemo } from 'react';
+
+export interface FeatureFlags {
+  lazyLoadPanelContent?: boolean;
+}
+
+export interface FeatureFlagContextProps {
+  featureFlags?: FeatureFlags;
+  children?: React.ReactNode;
+}
+
+const FeatureFlagContext = createContext<FeatureFlags>(undefined);
+
+const defaultFeatureFlagValues: FeatureFlags = {
+  lazyLoadPanelContent: false,
+};
+
+export const FeatureFlagContextProvider: FC<FeatureFlagContextProps> = ({
+  children,
+  featureFlags,
+}) => {
+  const ancestorFeatureFlags = useContext(FeatureFlagContext);
+
+  const currentContext = useMemo(
+    () => ({
+      ...(defaultFeatureFlagValues || {}),
+      ...(ancestorFeatureFlags || {}),
+      ...(featureFlags || {}),
+    }),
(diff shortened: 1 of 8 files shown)


gclark-eightfold:
I'm curious if this is the direction we want to go in for shipping Octuple changes faster.
I can see how feature flags can help land breaking changes into Octuple without massive efforts on regression testing, but I think we're also increasing our surface area for testing in the future.

Also, it might be a bit more complex to debug if component-level props are being ignored for context-level props.

If we decide we want to continue with feature flags though, I think we should:
1. Separating it out a bit more by component name might make it easier to follow & make it clear which feature flags pertain to look for if debugging is necessary.
2. Add JSDoc comments to each feature flag to indicate what it does / when to use it.
e.g.
```typescript
export interface FeatureFlags {
  panel?: {
    /**
     * If `true`, the panel content will render ONLY when `visible` is `true`.
     * Otherwise, the panel content will default to the `renderContentAlways` value.
     * 
     * < some information about when/why this should be used >
     */
    lazyLoadContent?: boolean;
  }
}
```

src/components/ConfigProvider/FeatureFlagProvider.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const ancestorFeatureFlags = useContext(FeatureFlagContext);
+
+  const currentContext = useMemo(
+    () => ({
+      ...(defaultFeatureFlagValues || {}),
+      ...(ancestorFeatureFlags || {}),
+      ...(featureFlags || {}),
+    }),
+    [featureFlags, ancestorFeatureFlags]
+  );
```
gclark-eightfold, line 33: Do we expect/support `featureFlags` changing between renders?
If this is just an application-level definition and not something loaded async from something like a backend config, we might be able to get a small optimization by wrapping it with a ref.

Also, not sure if this is a ES version-specific feature, but you can use the spread syntax on `undefined`
e.g.
```suggestion
  featureFlags,
}) => {
  const featureFlagsRef = useRef(featureFlags);
  const ancestorFeatureFlags = useContext(FeatureFlagContext);

  const currentContext = useMemo(
    () => ({
      ...defaultFeatureFlagValues,
      ...ancestorFeatureFlags,
      ...featureFlagsRef.current,
    }),
    [ancestorFeatureFlags]
  );
```

src/components/ConfigProvider/FeatureFlagProvider.tsx
```diff
@@ -0,0 +1,49 @@
...
+export interface FeatureFlagContextProps {
+  featureFlags?: FeatureFlags;
+  children?: React.ReactNode;
+}
+
+const FeatureFlagContext = createContext<FeatureFlags>(undefined);
+
+const defaultFeatureFlagValues: FeatureFlags = {
+  panelLazyLoadContent: false,
+};
```
gclark-eightfold, line 18: Should be able to initialize with the default values and remove the spreading of defaults to inner contexts.
```suggestion
const FeatureFlagContext = createContext<FeatureFlags>({
  panelLazyLoadContent: false,
});
```

src/components/ConfigProvider/FeatureFlagProvider.tsx
dcoblentz-eightfold: I had a fallback below where I did this: 
```
  useContext(FeatureFlagContext) || { ...defaultFeatureFlagValues };
```
This gave me a fallback with the default values if I run the code without the config provider being set for some reason. Maybe there's a better way to manage this though?

dcoblentz-eightfold: I would expect it to possibly be loaded async from a config, and possibly managed via gates, though I wouldn't expect it to change frequently at runtime.

jhoward-eightfold (approved):

src/components/ConfigProvider/FeatureFlagProvider.tsx
dcoblentz-eightfold: I ended up using the default values below, in cases where there's no config available

dcoblentz-eightfold:
:shipit:

PR #876, opened 2024-08-21
fix: make stepper region focusable if scrollable

cpatil-c-eightfold:
## SUMMARY:
In the case of onboarding flow, was getting axe devtools issue saying "Scrollable region must have keyboard access". So added a tabindex attribute to make the stepper region focusable to fix the issue.

https://eightfoldai.atlassian.net/browse/ENG-85455

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
image

- pull this locally and bring up storybook
- confirm that the stepper center region is focusable and user can scroll using the keyboard or not.
- confirm that other variants of the stepper component are still working as expected too.


src/components/Stepper/Stepper.tsx
```diff
@@ -522,6 +522,7 @@ export const Stepper: FC<StepperProps> = React.forwardRef(
                   className={styles.stepsContainer}
                   onScroll={handleScroll}
                   ref={stepsContainerRef}
+                  tabIndex={0}
```
jhoward-eightfold, line 525: seems like you could do this dynamically such that it isn't in the tab order when the scroll isn't present. something like (please test)

```suggestion
                  tabIndex={mergedScrollable ? 0 : null}
```

src/components/Stepper/Stepper.tsx
jhoward-eightfold, line 1: please add a test case that covers this scenario.

src/components/Stepper/Stepper.tsx
cpatil-c-eightfold: Updated tabindex attribute condition as suggested.

src/components/Stepper/Stepper.tsx
cpatil-c-eightfold: added a test case for scrollable stepper

jhoward-eightfold (approved):

PR #881, opened 2024-08-26
fix: select: address selected items getting unselected when expanded.

ysarvaiya-c-eightfold:
## SUMMARY:
This fix addresses the issue of Selected list item getting unselected when expanded or focus is on select dropdown container.

when keydown is called when enter is pressed on select dropdown. onClick event listener for close button on the selected pill item is getting called causing the selected item to remove from the selected list.

Fixed the issue by adding if guard to call toggle function only when currently active element in document is contained to the target close button.

https://eightfoldai.atlassian.net/browse/ENG-95192

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
- Pull the PR branch locally and run `yarn storybook`. 
- Verify the new added Form multiple select story on form and confirm selected items are not getting removed after expanding the select dropdown. (`http://localhost:2022/?path=/story/form--form-select-multiple`)

https://github.com/user-attachments/assets/a7c3ac19-7e23-4e8a-9307-e0769a4fa5d0

### DIFF
--- a/src/components/Form/Form.stories.tsx
+++ b/src/components/Form/Form.stories.tsx
@@ -59,125 +59,6 @@ export default {
   },
 } as ComponentMeta<typeof Form>;
 
-const Form_Select_Multiple_Story: ComponentStory<typeof Form> = (args) => {
-  const [form] = Form.useForm();
-  const [selected, setSelected] = useState<string>();
-  const layout = {
-    labelCol: { span: 1 },
-    wrapperCol: { span: 11 },
-  };
-  const actionsLayout = {
-    wrapperCol: { offset: 1, span: 11 },
-  };
-  const defaultOptions: SelectOption[] = [
-    {
-      text: 'Foo',
-      value: 'foo',
-    },
-    {
-      text: 'Bar',
-      value: 'bar',
-    },
-    {
-      text: 'Other',
-      value: 'other',
-    },
-  ];
-
-  const onSelectChange = (options: SelectOption[]) => {
-    if (!options) {
-      return onReset();
-    }
-    switch (options[0]) {
-      case 'foo' as unknown as SelectOption:
-        form.setFieldListValues({
-          note: 'Hello, Foo!',
-          selectValue: 'foo',
-        });
-        setSelected('foo');
-        return;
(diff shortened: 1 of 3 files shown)


src/components/Select/Select.tsx
```diff
@@ -577,7 +578,11 @@ export const Select: FC<SelectProps> = React.forwardRef(
             disabled={mergedDisabled}
             key={`select-pill-${index}`}
             label={opt.text}
-            onClose={() => toggleOption(opt)}
+            onClose={(event) => {
```
jhoward-eightfold, line 581: can you add unit test for the scenario?

src/components/Select/Select.tsx
```diff
@@ -577,7 +578,11 @@ export const Select: FC<SelectProps> = React.forwardRef(
...
+              if (contains(document.activeElement, event.target as Node)) {
```
jhoward-eightfold, line 582: whats the idea here? why does it toggle option on close at all? seems like onclick/onkeydown should be solely responsible for toggling an option

src/components/Select/Select.tsx
jhoward-eightfold, line 1: can you add a unit test for this behavior please

src/components/Select/Select.tsx
jhoward-eightfold: I only just now realized that this is the pill onClose and not the dropdown on close... is there a possibly a propagation issue in play here?

src/components/Form/Form.stories.tsx
jhoward-eightfold, line 1: pulled this locally and played around with it. It is having the desired outcoming, though I don't full understand why the bug is happenning

jhoward-eightfold (approved):

src/components/Select/Select.tsx
ysarvaiya-c-eightfold: No. I’ve debugged the propagation part. event propagation is not affecting the onClose behavior for the pill component.

jhoward-eightfold: Do you have an understanding of how this is getting triggered?

PR #899, opened 2024-10-24
a11y: Accessibility fix for floating dropdown

rmotwani-c-eightfold:
## SUMMARY:
Added Tab event for option selection, Focus border and props for Floating ui to maintain focus.

Jira task - https://eightfoldai.atlassian.net/browse/ENG-113537

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:

- With NVDA on, go to Select component and add dropDown props with "portal = true".
- use arrow keys to navigate between option and observe if there is a focus border visible.
- Use Enter, Space and Tab to select the option.


src/components/Dropdown/Dropdown.tsx
```diff
@@ -423,7 +427,9 @@ export const Dropdown: FC<DropdownProps> = React.memo(
           {getReference()}
           <ConditionalWrapper
             condition={portal}
-            wrapper={(children) => <FloatingPortal>{children}</FloatingPortal>}
+            wrapper={(children) => (
+              <FloatingPortal preserveTabOrder>{children}</FloatingPortal>
```
szhang-eightfold, line 431: this will help portal component stay in current tab order? If so could you add it to all places using FloatingPortal?

szhang-eightfold: seems like this is already default true based on doc https://floating-ui.com/docs/floatingportal, is this doing anything?

rmotwani-c-eightfold: I cannot say if its doing anyhting because I could not verify this in storybook. This seem to be a viable fix so i added it.  Was not aware that it is true by default.

szhang-eightfold (approved):

PR #909, opened 2024-11-12
fix: dropdown auto click the element when TAB operation is invoked

sravichandran-eightfold:
## SUMMARY:
Dropdown element is clicking on first element since instead of focus, we were calling onclick event

https://eightfoldai.atlassian.net/browse/ENG-116631

## CHANGE TYPE:

- [X] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

- [X] Tests for this change already exist

## TEST PLAN:
1. Go to storybook
2. Go to Dropdown, Single Dropdown
3. Use Keyboard to Tab and when you tab multiple times it should not close or click on the element

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -296,7 +296,7 @@ export const Dropdown: FC<DropdownProps> = React.memo(
           }, NO_ANIMATION_DURATION);
           const menuButtonEvent: HTMLButtonElement =
             document.activeElement as HTMLButtonElement;
-          menuButtonEvent.click();
+          menuButtonEvent?.focus?.();
         }
         if (event?.key === eventKeys.TAB && event.shiftKey) {
           timeout && clearTimeout(timeout);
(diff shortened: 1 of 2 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -296,7 +296,7 @@ export const Dropdown: FC<DropdownProps> = React.memo(
           }, NO_ANIMATION_DURATION);
           const menuButtonEvent: HTMLButtonElement =
             document.activeElement as HTMLButtonElement;
-          menuButtonEvent.click();
+          menuButtonEvent?.focus?.();
```
szhang-eightfold, line 299: I thought we use Tab to exit dropdown and also select the element we last focused in? Maybe worth checking with Swapnill too, but this looks fine to me

szhang-eightfold (approved):

src/components/Dropdown/Dropdown.tsx
sravichandran-eightfold: There is a escape key handling for this as well

PR #927, opened 2024-12-24
fix: added aria-label to the date range picker

nsanap-c-eightfold:
## SUMMARY:

- The aria-label was applied only to the wrapper div element, representing the entire group.
- Inside the wrapper div, there are two input fields: one for the start date and one for the end date.
- To improve accessibility, we added support for custom aria-label attributes for both input fields.
- Introduced two new props: `startDateInputAriaLabel` and `endDateInputAriaLabel`, allowing child components to pass custom `aria-label` values for the respective inputs.
- If no `aria-label` is provided through the new props, the aria-label attributes will not be added to the date range picker inputs by default.

https://eightfoldai.atlassian.net/browse/ENG-122300

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:

- Ensure that the aria-label props can be passed to the Date Range Picker component.
- Verify that the aria-label values are correctly rendered in the DOM.
- Inspect the DOM to confirm that the aria-label attributes are applied to the start and end date input fields.
- Verify that the aria-label props are visible and correctly displayed in Storybook.
- Ensure the Date Range Picker component is functioning properly and the aria-label values are accessible and applied as expected.

### DIFF
--- a/src/components/DateTimePicker/DatePicker/DatePicker.stories.tsx
+++ b/src/components/DateTimePicker/DatePicker/DatePicker.stories.tsx
@@ -224,35 +224,11 @@ const { RangePicker } = DatePicker;
 const Range_Picker_Story: ComponentStory<typeof RangePicker> = (args) => {
   return (
     <Stack direction="vertical" flexGap="m">
-      <RangePicker
-        {...args}
-        startDateInputAriaLabel={args.startDateInputAriaLabel}
-        endDateInputAriaLabel={args.endDateInputAriaLabel}
-      />
-      <RangePicker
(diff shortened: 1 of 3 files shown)


cpatil-c-eightfold (approved):

cpatil-c-eightfold:
LGTM

https://github.com/user-attachments/assets/3c721874-cf68-4530-b9e8-5cd82e9ddf12

szhang-eightfold (approved):

PR #928, opened 2024-12-27
Cpatil/added rest props for checkbox and radio components

cpatil-c-eightfold:
## SUMMARY:
- Added condition for role attribute in checkboxgroup and radiogroup components
- Added rest props in checkbox and radiobutton components

https://eightfoldai.atlassian.net/browse/ENG-122365

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Added rest props in checkbox and radiobutton components. So test if we can add attributes like `aria-required` onto the input element of these 2 components or not.

As we can see in the below screenshot, the `aria-required` is applied to all the checkbox input elements when we pass it through the items array and is also read as required by voice over, if it is true.
image

### DIFF
--- a/src/components/CheckBox/CheckBoxGroup.tsx
+++ b/src/components/CheckBox/CheckBoxGroup.tsx
@@ -107,7 +107,7 @@ export const CheckBoxGroup: FC<CheckboxGroupProps> = React.forwardRef(
       <div
         className={checkboxGroupClassNames}
         style={style}
-        role={items.length > 1 ? 'group' : undefined}
+        role={items?.length > 1 ? 'group' : undefined}
         aria-label={ariaLabel}
         id={id}
         ref={ref}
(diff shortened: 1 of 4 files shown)


szhang-eightfold (commented):
have you tested in octuple storybook that it is working? If so could you show screenshot or video of that

cpatil-c-eightfold:
Ok I have tested and add the screenshot in the Test plan section now @szhang-eightfold

src/components/CheckBox/CheckBoxGroup.tsx
```diff
@@ -107,7 +107,7 @@ export const CheckBoxGroup: FC<CheckboxGroupProps> = React.forwardRef(
       <div
         className={checkboxGroupClassNames}
         style={style}
-        role="group"
+        role={items.length > 1 ? 'group' : undefined}
```
sravichandran-eightfold, line 110: items?.length

sravichandran-eightfold (approved):

src/components/CheckBox/CheckBoxGroup.tsx
cpatil-c-eightfold: done

PR #930, opened 2025-01-06
fix: dropdown: removed outer container tabindex

krane-c-eightfold:
## SUMMARY:
as role of overlay container is generic but it may not be focusable at all times

changes- so added prop to set tabindex generically as well

N/A

https://eightfoldai.atlassian.net/browse/ENG-99020

## CHANGE TYPE:
- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
CASE: if overlay container is passed to component and  its role is set as presentation then container should not be focusable only options should be focusable

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -75,6 +75,7 @@ export const Dropdown: FC<DropdownProps> = React.memo(
         trigger = 'click',
         visible,
         width,
+        overlayWrapperTabIndex = 0,
       },
       ref: React.ForwardedRef<DropdownRef>
     ) => {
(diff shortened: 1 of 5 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -392,7 +392,6 @@ export const Dropdown: FC<DropdownProps> = React.memo(
               ref={refs.setFloating}
               style={dropdownStyles}
               className={dropdownClasses}
-              tabIndex={0}
```
sravichandran-eightfold, line 395: assign this to a prop `tabIndex` and then set the value in primary repo

cpatil-c-eightfold (approved):

src/components/Dropdown/Dropdown.tsx
```diff
@@ -75,6 +75,7 @@ export const Dropdown: FC<DropdownProps> = React.memo(
         trigger = 'click',
         visible,
         width,
+        overlayWrapperTabIndex = 0,
```
sravichandran-eightfold, line 78: The property name is too long, can we keep it simple

krane-c-eightfold: made it overlayTabIndex

sravichandran-eightfold (approved):

Issue #942, opened 2025-02-05
a11y: accessibility attributes not respected for select options.


ptiwari-eightfold:
Accessibility attributes for select options aren't deduced correctly resulting in following a11y issues.

<img width="1396" alt="Image" src="https://github.com/user-attachments/assets/72bee192-734c-4a94-a0b8-e00cced8486d" />

<img width="1396" alt="Image" src="https://github.com/user-attachments/assets/3a8f9068-7692-4bff-bed4-3ef952ebd42e" />

<img width="1396" alt="Image" src="https://github.com/user-attachments/assets/5170152e-0f7b-481b-856d-00927f757360" />

PR #943, opened 2025-02-05
fix: a11y: Fix role inheritance for select options

ptiwari-eightfold:
## SUMMARY:
Roles for select options were set on internal elements enclosed inside the li element. This resulted li interfering with accessibility tools while deducing the aria attributes for the enclosed items as all accessibility were set on enclosed items.

This commit sets the role for enclosing li as presentation.

`role="presentation"` (or equivalently `role="none"`) is an ARIA attribute that indicates an element is being used only for visual presentation and should be ignored by assistive technologies like screen readers. It effectively removes the semantic meaning of the element while keeping its visual appearance and behaviour.

**Before**
Image

Image

Image

**After**
image

This helps in deducing the aria attributes for the enclosed items correctly.
https://github.com/EightfoldAI/octuple/issues/942
https://eightfoldai.atlassian.net/browse/ENG-129109
## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

### DIFF
--- a/src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
+++ b/src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
@@ -170,6 +170,10 @@ export const MenuItemButton: FC<MenuItemButtonProps> = forwardRef(
       return dropdownMenuItems ? dropdownMenuButton() : menuButton();
     };
 
-    return <li className={menuItemClassNames}>{renderedItem()}</li>;
+    return (
+      <li className={menuItemClassNames} role="presentation">
+        {renderedItem()}
+      </li>
+    );
   }
 );
--- a/src/components/Menu/MenuItem/MenuItemCustom/MenuItemCustom.tsx
+++ b/src/components/Menu/MenuItem/MenuItemCustom/MenuItemCustom.tsx
@@ -21,7 +21,7 @@ export const MenuItemCustom: FC<MenuItemCustomProps> = forwardRef(
     ]);
 
     return (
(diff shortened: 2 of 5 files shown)


ychhabra-eightfold (approved):

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
```diff
@@ -170,6 +170,10 @@ export const MenuItemButton: FC<MenuItemButtonProps> = forwardRef(
       return dropdownMenuItems ? dropdownMenuButton() : menuButton();
     };
 
-    return <li className={menuItemClassNames}>{renderedItem()}</li>;
+    return (
+      <li className={menuItemClassNames} role="presentation">
```
ychhabra-eightfold, line 174: should we only do this for select ? menu is used in dropdowns as well

ychhabra-eightfold (changes requested):

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
ptiwari-eightfold: Since all the props (even accessibility ones) are set on the enclosed elements. Presence of li elements will interfere in a similar manner even for other cases, because someone trying to set the aria attributes would assume they are set in the MenuItems, but instead they are being set one component level inside.

ychhabra-eightfold (approved):

PR #975, opened 2025-04-16
fix: remove non existent tooltip aria-describedby reference

uagarwal-eightfold:
## SUMMARY:
A11y fix: 
Only add `aria-describedby` on a tooltip reference element when the tooltip is visible. When the tooltip is not visible the tooltip element is not in DOM and hence the reference is broken.

https://eightfoldai.atlassian.net/browse/ENG-141226

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
1. Go to storybook for multi select. /?path=/story/select--multiple
2. Select the long text item which will be showing tooltips.
3. Verify that there is no ARC toolkit error for invalid aria-describedby.
4. Hover over a pill to show its tooltip, and while the tooltip is visible, verify that aria-describedby is set on the reference element.
5. [Regression test]: Verify that tooltip hover in/out works fine.

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -397,15 +397,17 @@ export const Tooltip: FC<TooltipProps> = React.memo(
               'tooltip-reference',
             ]);
 
+            const ariaDescribedBy =
+              node.props?.['aria-describedby'] || tooltipId?.current;
+
             const clonedElementProps: RenderProps = {
               id: node.props?.id ? node.props?.id : tooltipReferenceId?.current,
               key: node.props?.key ? node.props?.key : tooltipId?.current,
               // If the content is not a string, the element of the content should be
               // manually targeted by id via `aria-describedby` to be announced by screen readers.
               // As this is an edge case, don't worry about it here, instead take the override if available.
-              'aria-describedby': node.props?.['aria-describedby']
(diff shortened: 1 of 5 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -403,9 +403,12 @@ export const Tooltip: FC<TooltipProps> = React.memo(
               // If the content is not a string, the element of the content should be
               // manually targeted by id via `aria-describedby` to be announced by screen readers.
               // As this is an edge case, don't worry about it here, instead take the override if available.
-              'aria-describedby': node.props?.['aria-describedby']
+              // Only assign aria-describedby if the tooltip element is present in the DOM.
+              'aria-describedby': mergedVisible
                 ? node.props?.['aria-describedby']
-                : tooltipId?.current,
+                  ? node.props?.['aria-describedby']
+                  : tooltipId?.current
```
bakcay-eightfold, line 410: Can we simplify this nested ternary operations

uagarwal-eightfold: I had tried to add parantheses, but eslint removed them. Will check another way.

uagarwal-eightfold: Fixed

PR #979, opened 2025-04-22
fix: snackbar: fixed focus for info bar close button

rmotwani-c-eightfold:
## SUMMARY:

- Previously, focus was incorrectly moving to the info bar close button even when it wasn’t needed.
- Introduced a new prop, moveFocusToCloseButton, to ensure focus only shifts to the close button when the info bar is rendered via the Snackbar component.

NA

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:

- pull branch down locally and start storybook (yarn storybook)
- go to Infobar.
- Verify if focus does not move to the close button for the infobar.
- go to Snackbar
- Set closable to true in the controls.
- Verify if focus goes to the close button on snackbar.

https://github.com/user-attachments/assets/5047b174-d3ff-48e3-8bd4-8fc156523fcb


src/components/Snackbar/Snackbar.types.ts
```diff
@@ -30,6 +30,15 @@ export interface SnackbarProps extends Omit<InfoBarsProps, 'onClick'> {
    * @default top-center
    */
   position?: SnackbarPosition;
+  /**
+   * Last focusable element of the snackbar
+   */
+  lastFocusableElement?: HTMLElement;
```
ypatadia-eightfold, line 36: is this being used?

ypatadia-eightfold (changes requested):

src/components/Snackbar/Snackbar.types.ts
rmotwani-c-eightfold: yes this is used to get the last focusable item from where ever the snack bar is called from

ypatadia-eightfold (approved):

PR #982, opened 2025-04-23
Added handlekeydown event in panel component to close panel on escape key

cgarge-c-eightfold:
## SUMMARY:
Added handlekeydown event in panel component to close panel on escape key

https://eightfoldai.atlassian.net/browse/ENG-140714

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Open Panel Storybook 
URL : ?path=/story/panel--header-actions 
Open panel 
Click "Escape" / "ESC" key , panel should close 

https://github.com/user-attachments/assets/42709301-ae52-4fdc-afac-a755da8f8203


src/components/Panel/Panel.tsx
```diff
@@ -137,6 +138,19 @@ export const Panel = React.forwardRef<PanelRef, PanelProps>(
...
+        onClose?.(event);
+      }
+    };
+
+    useLayoutEffect(() => {
+      document.addEventListener('keydown', handleEscapeKey);
+      return () => {
+        document.removeEventListener('keydown', handleEscapeKey); 
+      };
+    }, [handleEscapeKey, visible]);
```
samyakj-eightfold, line 152: Why do we need this dependancy array ?

cgarge-c-eightfold: Can remove **handleEscapeKey** from dependency array , 
need to keep **visible** as we are utilizing it in  **handleEscapeKey**

samyakj-eightfold: I did not get your point here: `need to keep visible as we are utilizing it in handleEscapeKey` ?
The dependancy array is not based on what is utilised or not, it should be based on what re-triggers the side effect
Did we try removing visible from dependancy array? What was the observation here ?

samyakj-eightfold:
I see some risks in this PR (I would suggest to get this reviewed by octuple team):
1. Can we optimise this further ? We should only add the event listener to the document when the panel is visible
2. What if there is an existing keydown event listener on document, do you also remove those when you remove the escape event listener

src/components/Panel/Panel.tsx
```diff
@@ -137,6 +138,19 @@ export const Panel = React.forwardRef<PanelRef, PanelProps>(
       );
     }, [mergedLocale]);
 
+    const handleEscapeKey = (event: KeyboardEvent) => {
+      if (event.key === 'Escape' && visible ) {
+        onClose?.(event);
+      }
+    };
+
+    useLayoutEffect(() => {
```
ypatadia-eightfold, line 147: why are we using useLayoutEffect and not useEffect?

ypatadia-eightfold (changes requested):

src/components/Panel/Panel.tsx
cgarge-c-eightfold: We are adding event listener on dom directly 
hence using useLayoutEffect

ypatadia-eightfold: why can't we add it to useEffect?

ypatadia-eightfold (changes requested):

src/components/Panel/Panel.tsx
ypatadia-eightfold: why can't we add it to useEffect?

cgarge-c-eightfold: Updated code

src/components/Panel/Panel.tsx
```diff
@@ -137,6 +137,19 @@ export const Panel = React.forwardRef<PanelRef, PanelProps>(
       );
     }, [mergedLocale]);
 
+    const handleEscapeKey = (event: KeyboardEvent) => {
```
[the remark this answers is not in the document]
cgarge-c-eightfold: Fixed

ypatadia-eightfold (approved):

PR #987, opened 2025-04-29
fix: multi-select label voice over a11y

ypatadia-eightfold:
## SUMMARY:
Screen reader will announce whether it's select or multiselect

https://eightfoldai.atlassian.net/browse/ENG-136017

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
1. /?path=/story/select--multiple
2. turn on voice mode
3. Focus on select box
4. Make sure voice over is announcing multi-select
image


src/components/Select/Select.tsx
```diff
@@ -883,6 +883,10 @@ export const Select: FC<SelectProps> = React.forwardRef(
                   ref={inputRef}
                   aria-activedescendant={currentlySelectedOption.current?.id}
                   aria-controls={selectMenuId?.current}
+                  aria-expanded={dropdownVisible}
+                  aria-multiselectable={multiple}
+                  title={placeholder + (multiple ? ' multi-select' : '')}
```
uagarwal-eightfold, line 888: Can we provide a way to override from caller app to be able to support i18n?

ypatadia-eightfold: we can just keep title={placeholder} and you can pass the placeholder with i18n

uagarwal-eightfold: We can override this from individial apps using textInputProps, so this change is good.

uagarwal-eightfold (approved):

PR #1113, opened 2026-03-31
fix(Dropdown): updated structure for dropdown menu list item

ssherigar-c-eightfold:
## SUMMARY:
**MenuItemButton**
  - Moved interactivity from button → li (with role, tabIndex, keyboard handlers)
  - Added full keyboard support (Arrow keys, Enter/Space)
  - Simplified nested items (use span instead of button)

**Dropdown**
  - Removed default referenceRole='button'
  - Removed incorrect role='listbox' from overlay
  - Fixed aria-haspopup usage 
      - The fix removes the default value 'true' from ariaHaspopupValue:                                                                                                                                                                                                                
          _Before_ : ariaHaspopupValue defaulted to 'true', so aria-haspopup="true" was always present on the dropdown button, even when no value  was passed.                                                                                                                                                                         
            _After_ : ariaHaspopupValue has no default. When nothing is passed (or false is explicitly passed), aria-haspopup is omitted entirely. The attribute only appears when a string value is explicitly provided by the parent component. 
            
  - Role only applied when explicitly passed

**List**
  - Added ariaLabelledBy → maps to aria-labelledby, to allow lists to be programmatically associated with visible labels
  - Only render the role attribute for ul and li tag when explicitly provided by the parent. Remove default/implicit role assignment to avoid unnecessary roles in the DOM.

NA
https://eightfoldai.atlassian.net/browse/ENG-167674
https://eightfoldai.atlassian.net/browse/ENG-167852
https://eightfoldai.atlassian.net/browse/ENG-167512


## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Check dropdown list items for accessibility compliance
**Dropdown**
- Verify click and keyboard (Enter/Space) interactions work correctly
- Ensure arrow keys support proper navigation between items
- Confirm button is removed and replaced with a non-interactive element (e.g., span), with interaction handled at the parent level
- Validate that appropriate roles are applied
- Ensure correct use of aria-haspopup, aria-expanded, and aria-controls
- Confirm existing ARIA attributes are preserved
- Ensure no incorrect roles are applied to the overlay


ychhabra-eightfold (commented):
## Regression Risk Analysis

This PR restructures Dropdown menu list items for a11y — moving interactivity from inner `<button>` to `<li>`, fixing ARIA roles, and adding keyboard navigation. The a11y intent is solid, but there are several areas with regression risk that need attention.

**Overall assessment:** The structural change from `<li role="presentation"> > <button role="menuitem">` to `<li role="menuitem">` is directionally correct for ARIA compliance, but the scope of changes across Dropdown, Menu, Select, and multiple snapshot-dependent components creates a wide blast radius. See inline comments for specific risks.

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
```diff
@@ -111,30 +102,30 @@ export const MenuItemButton: FC<MenuItemButtonProps> = forwardRef(
...
-            ref={ref}
-            role={role}
-          >
+          <span className={styles.menuOuterButton}>
             {iconProps && alignIcon === MenuItemIconAlign.Left && getIcon()}
             <span className={styles.menuItemWrapper}>
               <span className={styles.itemText}>
                 <span className={styles.label}>{text}</span>
               </span>
             </span>
```
ychhabra-eightfold: **Regression Risk: `...rest` spread moved from `<button>` to `<li>`**

Previously, `...rest` was spread onto the inner `<button>`. Now it's on the `<li>`. This is a **breaking change** for consumers who pass HTML button-specific attributes (e.g., `type`, `form`, `formAction`) or event handlers expecting `HTMLButtonElement` targets.

The type was also changed from `OcBaseProps<HTMLButtonElement>` to `OcBaseProps<HTMLLIElement>`, which is correct for TypeScript consumers, but any JavaScript consumers passing button-specific props will get silent runtime bugs.

Worth calling out as a breaking change in release notes.

src/components/Menu/Menu.tsx
```diff
@@ -89,7 +89,7 @@ export const Menu: FC<MenuProps> = ({
   const getHeader = (): JSX.Element =>
     header && (
       <div className={headerClassNames}>
-        <div className={styles.heading}>{header}</div>
```
ychhabra-eightfold: **Regression Risk: Hardcoded `<h2>` may break heading hierarchy**

Changing from `<div>` to `<h2>` is good for a11y (gives the menu a semantic heading), but hardcoding `<h2>` may break the document heading hierarchy depending on where the Menu is rendered. If a Menu appears inside a section that's already under an `<h2>`, this creates an incorrect heading level.

Consider making the heading level configurable (e.g., `headerLevel` prop defaulting to `h2`), or use `role="heading" aria-level={n}` for flexibility.

src/components/Select/Select.test.tsx
```diff
@@ -609,10 +609,7 @@ describe('Select', () => {
     fireEvent.keyDown(select, { key: 'Enter' });
     const listbox = await waitFor(() => getAllByRole('option'));
     expect(listbox).toHaveLength(options.length);
-    await waitFor(() =>
-      expect(screen.getByTestId('option1-test-id').matches(':focus')).toBe(true)
-    );
-    expect(screen.getByTestId('option1-test-id').matches(':focus')).toBe(true);
+    expect(select).toHaveFocus();
   });
```
ychhabra-eightfold: **Regression Risk: Focus tests weakened rather than fixed**

Multiple tests previously asserted that opening the dropdown focused the first option (`option1-test-id.matches(':focus')`). These have been replaced with `expect(select).toHaveFocus()`, meaning the **select input retains focus** instead of moving to the options.

This is a significant UX/a11y behavior change:
- **Before:** Opening a Select with keyboard moved focus into the listbox options, allowing immediate arrow key navigation of the options list.
- **After:** Focus stays on the input, meaning users must use a different mechanism to navigate options.

If this is intentional (virtual focus via `aria-activedescendant`), that pattern requires `aria-activedescendant` to be set during arrow key navigation — but the updated filterable test only checks `aria-activedescendant` **after a click**, not during keyboard nav. This may leave keyboard-only users without proper option announcement.

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
```diff
@@ -91,17 +91,8 @@ export const MenuItemButton: FC<MenuItemButtonProps> = forwardRef(
       [MenuSize.small, IconSize.Small],
     ]);
 
-    const menuButton = (isNested: boolean = false): JSX.Element => (
-      <button
-        className={styles.menuItemButton}
-        disabled={disabled}
-        tabIndex={tabIndex}
-        type={htmlType}
-        {...rest}
```
ychhabra-eightfold: **Regression Risk: `secondaryButton` lost its click handler and ref**

The inner `<button>` in `secondaryButton()` was replaced with a `<span>`, but the `onClick={handleOnClick}` and `ref={ref}` that were on that button are now gone. The `<li>` has `onClick={!dropdownMenuItems ? handleOnClick : undefined}`, but `secondaryButton` items may or may not have `dropdownMenuItems`.

If a menu item uses `secondaryButtonProps`, the primary text area (`menuOuterButton` span) is no longer independently clickable — clicks only work on the `<li>`. This might be fine, but verify it doesn't break split-button menu items where the primary and secondary actions are distinct.

src/components/Select/Select.test.tsx
ssherigar-c-eightfold: Fixed

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
ssherigar-c-eightfold: The type was updated from OcBaseProps<HTMLButtonElement> to                   OcBaseProps<HTMLLIElement>, so TypeScript consumers will get a compile-time error if they pass button-specific props like type, form, or formAction.

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
```diff
@@ -111,30 +102,30 @@ export const MenuItemButton: FC<MenuItemButtonProps> = forwardRef(
...
             {iconProps && alignIcon === MenuItemIconAlign.Left && getIcon()}
             <span className={styles.menuItemWrapper}>
               <span className={styles.itemText}>
                 <span className={styles.label}>{text}</span>
               </span>
             </span>
             {iconProps && alignIcon === MenuItemIconAlign.Right && getIcon()}
-          </button>
+          </span>
           <span className={styles.menuInnerButton}>
```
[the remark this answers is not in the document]
ssherigar-c-eightfold: Fixed, tabIndex={disabled ? -1 : active ? 0 : tabIndex}
Disabled items now get tabIndex={-1}, removing them from the tab sequence and arrow key navigation, matching the old native <button disabled> behavior

src/components/List/List.tsx
```diff
@@ -261,8 +278,8 @@ export const List = <T extends any>({
       {listType === 'ul' && (
         <ul
           id={id}
-          role={role}
-          aria-label={ariaLabel}
+          {...(role !== undefined && { role })}
+          aria-labelledby={ariaLabelledBy}
```
ychhabra-eightfold, line 283: Regression: the existing `aria-label` prop is destructured (line 39 `'aria-label': ariaLabel`) but no longer rendered. Replacing it with `aria-labelledby` silently breaks every consumer that passes `aria-label` for the list's accessible name — and they get no warning or migration path.

Keep both: render `aria-label={ariaLabel}` alongside `aria-labelledby={ariaLabelledBy}`. They're not mutually exclusive, and `aria-labelledby` should be additive.

src/components/List/List.tsx
```diff
@@ -273,8 +290,8 @@ export const List = <T extends any>({
       {listType === 'ol' && (
         <ol
           id={id}
-          role={role}
-          aria-label={ariaLabel}
+          {...(role !== undefined && { role })}
+          aria-labelledby={ariaLabelledBy}
```
ychhabra-eightfold, line 296: Same as the `<ul>` branch: dropping `aria-label` here is a silent breaking change for `<ol>` consumers. Render both `aria-label` and `aria-labelledby`.

src/components/Menu/Menu.tsx
```diff
@@ -25,7 +25,7 @@ export const Menu: FC<MenuProps> = ({
   onCancel,
   onChange,
   onOk,
-  role = 'menu',
+  role,
```
ychhabra-eightfold, line 28: Removing the `'menu'` default cascades into a real accessibility regression: `Menu` now passes `role={undefined}` to `List`, which (combined with the List-side `role` change) renders `<ul>` with no role at all — while its `<li>` children still carry `role='menuitem'`. Per WAI-ARIA, `menuitem` must have a `menu`/`menubar` ancestor. Visible in the updated `Menu.test.tsx.snap`: the `ul` has no role but every `li` is `role="menuitem"`.

Keep `role = 'menu'` as the default (or pass it explicitly to `List` here) to preserve the parent/child role contract.

src/components/Menu/Menu.tsx
```diff
@@ -89,7 +89,7 @@ export const Menu: FC<MenuProps> = ({
   const getHeader = (): JSX.Element =>
     header && (
       <div className={headerClassNames}>
-        <div className={styles.heading}>{header}</div>
+        <h2 className={styles.heading}>{header}</h2>
```
ychhabra-eightfold, line 92: Hard-coding `<h2>` forces a fixed heading level on every Menu instance regardless of the surrounding document outline. A menu inside an `<h3>` section, or one with no heading parent at all, will produce an invalid heading hierarchy.

Make the heading level configurable (e.g. a `headerAs`/`headerLevel` prop) and default to the previous `<div>` to keep existing renders backward-compatible. If the goal is just to make headers focusable/labellable, consider `role='heading' aria-level={n}` driven by a prop instead.

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
```diff
@@ -177,7 +182,16 @@ export const MenuItemButton: FC<MenuItemButtonProps> = forwardRef(
     };
 
     return (
-      <li className={menuItemClassNames} role={listItemRole ?? 'presentation'}>
+      <li
+        className={menuItemClassNames}
+        aria-disabled={disabled ? true : undefined}
```
ychhabra-eightfold, line 187: This breaks the roving-tabindex pattern that `menu`/`menubar` requires. `tabIndex` prop defaults to `0` (line 33), so every non-disabled item evaluates to `tabIndex=0` — meaning Tab stops on every menu item instead of just the active one. Confirmed in the new snapshots: non-disabled items all render `tabindex="0"`.

The correct shape is: `tabIndex = disabled ? -1 : (active ? 0 : -1)` with the `Menu`/`List` parent driving which item is currently the tab stop. Otherwise focus management inside the menu is broken for keyboard users.

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
```diff
@@ -177,7 +182,16 @@ export const MenuItemButton: FC<MenuItemButtonProps> = forwardRef(
...
+        tabIndex={disabled ? -1 : active ? 0 : tabIndex}
```
ychhabra-eightfold, line 188: `{...rest}` is spread *after* `aria-disabled` and `tabIndex` but *before* `onClick`/`onKeyDown`. Any consumer-supplied `aria-disabled`, `tabIndex`, `role`, etc. in `rest` will silently override the deliberately-set values above (and vice-versa for click/keydown). Move `{...rest}` to be the first prop on the element so explicit values always win, or split it explicitly.

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
```diff
@@ -177,7 +182,16 @@ export const MenuItemButton: FC<MenuItemButtonProps> = forwardRef(
...
+        {...rest}
```
ychhabra-eightfold, line 189: When `dropdownMenuItems` is set, click is intentionally disabled here, and `handleLiKeyDown` (line 152) also skips Enter/Space in that branch. The nested content is now a `<span>` (no native interactivity). That leaves the dropdown-trigger menu item with no keyboard activation path on the `<li>` itself — it relies entirely on the wrapping `Dropdown` reference handlers being attached to this same node.

Please confirm that the `menuRenderer`-wrapped Dropdown does in fact bind its trigger handlers to this `<li>` (and not to the inner `<span>`). If not, Enter/Space will no longer open nested dropdowns.

src/components/Menu/MenuItem/MenuItem.types.ts
```diff
@@ -68,7 +68,7 @@ export interface MenuItemProps {
   menuRenderer?: MenuRenderFunction;
 }
 
-type NativeMenuButtonProps = Omit<OcBaseProps<HTMLButtonElement>, 'children'>;
+type NativeMenuButtonProps = Omit<OcBaseProps<HTMLLIElement>, 'children'>;
```
ychhabra-eightfold, line 71: Public-API type change: `NativeMenuButtonProps` switched from `HTMLButtonElement` to `HTMLLIElement`. Any consumer that typed an event handler or `ref` against `MenuItemButtonProps` (e.g. `React.MouseEvent<HTMLButtonElement>`) will get a TS error. Worth calling out in the changelog as a breaking type change, even if the runtime behavior intent is the same.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -470,19 +469,23 @@ export const Dropdown: FC<DropdownProps> = React.memo(
...
           const currentAriaHaspopup =
             ariaRef.current.getAttribute('aria-haspopup');
-          if (currentRole !== 'combobox' && !currentAriaHaspopup) {
-            ariaRef.current.setAttribute('aria-haspopup', ariaHaspopupValue);
+          if (!currentAriaHaspopup) {
+            if (currentRole === 'combobox') {
+              ariaRef.current.setAttribute('aria-haspopup', 'listbox');
+            } else if (ariaHaspopupValue !== false) {
+              ariaRef.current.setAttribute('aria-haspopup', ariaHaspopupValue);
+            }
```
ychhabra-eightfold, line 481: Behavior change worth flagging in the description: previously combobox refs were left without `aria-haspopup` (the test asserted `.toBeNull()`); now they're force-set to `'listbox'`. ARIA 1.2 allows `'listbox' | 'tree' | 'grid' | 'dialog' | 'menu'` for combobox popups — `'listbox'` is the right default for `Select`, but a combobox whose popup is a `menu` or `dialog` will now get the wrong value with no way to opt out.

Consider honoring `ariaHaspopupValue` for combobox too (fall through to it when provided), so callers can override the implicit `'listbox'`.

ychhabra-eightfold (commented):
Review of accessibility refactor. Strong direction overall — moving interactivity onto `<li>` to drop redundant button wrappers is the right ARIA pattern for `menuitem`. Flagging a few regressions and concerns below.

**High-impact regressions**
1. `List` silently drops `aria-label` on the rendered `<ul>`/`<ol>`. The destructured prop `'aria-label': ariaLabel` is no longer applied — every existing consumer relying on `aria-label` loses its accessible name with no migration path. `aria-labelledby` should be additive, not a replacement.
2. `Menu` removed the default `role='menu'`, so the `<ul>` rendered by `List` no longer has a container role, while its children still claim `role='menuitem'`. Per WAI-ARIA, `menuitem` requires a `menu`/`menubar` ancestor — visible in the new snapshots (`<ul class="list-container vertical">` with `<li role="menuitem">` inside). This is a spec violation that screen readers will pick up on.
3. `MenuItemButton` sets `tabIndex={disabled ? -1 : active ? 0 : tabIndex}` where `tabIndex` defaults to `0`. That means every non-disabled menu item is `tabindex=0`, breaking the roving-tabindex pattern that menus require — Tab will now stop on every item. Confirmed in the new snapshots (all non-disabled items show `tabindex="0"`).

**Other concerns**
- `MenuItemCustom` and `Menu` overlay (`role='listbox'`) defaults were removed too — same shape of problem as #2 for any consumer that was relying on the implicit role.
- `MenuItemButton` spreads `{...rest}` on the `<li>` *after* `aria-disabled`/`tabIndex` but *before* `onClick`/`onKeyDown` — accidental `aria-disabled`/`tabIndex` keys in `rest` will silently override the deliberate values.
- The dropdown-trigger menu item (`dropdownMenuItems` branch) loses keyboard activation: `handleLiKeyDown` skips Enter/Space when `dropdownMenuItems` is set, and there's no `<button>` underneath anymore. Worth verifying the wrapping `Dropdown` reference handles this.
- `Menu` header was hard-coded to `<h2>`. That forces a heading level on every consumer regardless of where the menu lives in the document outline — should be configurable (e.g. `headerLevel` prop) and default to a non-heading element to preserve current behavior.
- `Dropdown` now eagerly sets `aria-haspopup="listbox"` on combobox elements (the test was updated from `toBeNull()` to `'listbox'`). This is correct per ARIA 1.2 but is a behavior change — confirm no consumer was deliberately leaving it off (e.g. for combobox-with-menu popup).
- `MenuItem.types.ts` changed `NativeMenuButtonProps` from `OcBaseProps<HTMLButtonElement>` to `OcBaseProps<HTMLLIElement>`. Public type change — breaks any consumer that destructured a button-typed ref/event from `MenuItemButtonProps`.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -563,7 +568,6 @@ export const Dropdown: FC<DropdownProps> = React.memo(
               onClick={handleDropdownClick}
               onKeyDown={handleFloatingKeyDown}
               id={dropdownId}
-              role={role}
               {...overlayProps}
```
ychhabra-eightfold, line 571: Removing the default `role='listbox'` on the overlay is reasonable, but combine it with the `referenceRole` default also being removed and you get a regression for any consumer that was relying on either default — no warning, just silently missing roles. Consider keeping the `role` prop pass-through (`{...(role !== undefined && { role })}`) so callers that explicitly pass one still see it applied.

src/components/Menu/MenuItem/MenuItemCustom/MenuItemCustom.tsx
```diff
@@ -21,7 +21,7 @@ export const MenuItemCustom: FC<MenuItemCustomProps> = forwardRef(
     ]);
 
     return (
-      <li className={menuItemClassNames} role={role}>
+      <li className={menuItemClassNames} {...(role !== undefined && { role })}>
```
ychhabra-eightfold, line 24: Same theme as the other default-removals: dropping the `role='presentation'` default leaves custom items with no role. Inside a `role='menu'` ul, a roleless `<li>` with arbitrary content will be exposed as a generic list item to AT, which may interfere with the menu's accessible structure. Either keep the `presentation` default or document why callers must now opt in.

src/components/Menu/Menu.tsx
ssherigar-c-eightfold: This is a requirement , we need to use a semantic h2, tag for the heading
styling is updated, as per previous design

src/components/Menu/Menu.tsx
ssherigar-c-eightfold: default role is removed from both ul and li tag, role will render only when explicitly provided by the parent

src/components/Menu/MenuItem/MenuItemCustom/MenuItemCustom.tsx
ssherigar-c-eightfold: default role is removed from both ul and li tag, role will render only when explicitly provided by the parent

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
ssherigar-c-eightfold: Tab cycling through all menu items is intentional — this component does not use the WAI-ARIA roving tabindex pattern. Each item remains a Tab stop to support both Tab and arrow key navigation, and shouldCloseOnTab provides a clear exit path. tabIndex=0 is intentional, so no change is needed

src/components/Menu/MenuItem/MenuItemButton/MenuItemButton.tsx
ssherigar-c-eightfold: Enter/Space for nested dropdowns still works via getItemProps()’s merged onKeyDown. handleLiKeyDown only triggers currentTarget.click() when !dropdownMenuItems; otherwise, the event falls through and Floating UI handles submenu activation. Verified in practice: arrow key navigation and Enter/Space both work correctly, so no change is needed.

src/components/Dropdown/Dropdown.tsx
ssherigar-c-eightfold: Removed the redundant role from the wrapper div, since the same role is already correctly applied to the ul.

ychhabra-eightfold (approved):

PR #1131, opened 2026-06-04
Exclude sub header options from the Select results count announcement

shubham-kumar1-eightfold:
## SUMMARY:

The `Select` live region announced an inflated result count for grouped dropdowns because presentational sub header options (`role="presentation"`) were counted as selectable results. A grouped, filterable `Select` with 3 options under 2 group headers announced *"5 matches found."* to screen readers — and announced a false positive count when only headers remained visible after filtering.

- Exclude `MenuItemType.subHeader` options from the visible-options count used to build the live region message.
- No visual or DOM change — only the announced count is affected, and flat (ungrouped) `Select`s are completely unaffected (options without a `type` are counted exactly as before).
- Also corrects the no-results case: when a query leaves only group headers visible, the live region now correctly announces *"No match found for &lt;query&gt;"*.

**Note:** Fixes the issue shown in the video of the latest comment in the attached JIRA

N/A

https://eightfoldai.atlassian.net/browse/ENG-167727

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:

**Automated**

1. `yarn jest src/components/Select/Select.test.tsx` — all 63 Select tests pass, including the new `Live region announcements › Grouped options with sub headers` suite (4 cases: count on open, count while filtering, single-match, and headers-only no-match). Verified red→green: the 4 new tests fail against the unpatched component and pass with the fix.

**Manual (screen reader)**
Test groupID: demo@eightfolddemo-pasthana.com
 
1. Open the dropdown with VoiceOver / NVDA active.
2. Go to homepage -> Create position -> Duplicate from position -> Duplicate on any position -> Start typing, eg. 's'
3. Confirm the live region announces only the number of selectable options (e.g. *"3 matches found."*) and no longer counts the group headers.
4. Type a query that matches a subset of options; confirm the announced count reflects only matching options.
5. Type a query that matches nothing; confirm it announces *"No match found for &lt;query&gt;"* rather than a header-only count.

### DIFF
--- a/src/components/Select/Select.test.tsx
+++ b/src/components/Select/Select.test.tsx
@@ -4,6 +4,7 @@ import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
 import MatchMediaMock from 'jest-matchmedia-mock';
 import { SelectShape, SelectSize } from './Select.types';
 import { Select } from './';
+import { MenuItemType } from '../Menu';
 import { sleep } from '../../tests/Utilities';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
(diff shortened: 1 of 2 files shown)


schakka-eightfold (approved):

agupta-eightfold (approved):
lg

PR #1140, opened 2026-07-02
fix(Table): add scroll tolerance so Scroller arrows reach the last column at fractional zoom

ayush-raj-eightfold:
## SUMMARY:
The horizontal scroll **arrow buttons** on `<Table showScroller>` can't scroll all the way to the last columns at browser zoom below 100% (e.g. 90%). The bottom scrollbar/slider works fine — only the arrow buttons fall short.

### Root cause
The `Scroller` builds its scroll stop-points from the **declared** column widths (e.g. 128px), advancing with `scrollOffsets.find(offset => offset > scrollLeft)`. At fractional zoom the browser snaps each 128px column to a slightly smaller rendered value (~127.78px at 90% / DPR 1.8) while the stop-points stay integer. So after a click targeting 128 the body settles at 127.78, and the next click's `find(offset => offset > 127.78)` returns **128 again** → scrolling snaps right back → the arrow is stuck. At 100% zoom rendered width == declared width, which is why it only reproduces below 100%.

This is a generic Octuple bug affecting any table using `showScroller` with fixed-pixel columns.

### Fix
Add a 1px tolerance in `Scroller`'s `onClick` so the comparison skips the
boundary the scroller is already snapped to:
- right arrow: `offset > scrollLeft + 1`
- left arrow: `offset < scrollLeft - 1`
- (same for both RTL branches)

https://eightfoldai.atlassian.net/browse/ENG-199352

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
Tested locally -

https://github.com/user-attachments/assets/03f2c983-4989-49ae-bcf1-2f6dbec58fb2

### DIFF
--- a/src/components/Table/Internal/Body/Scroller.tsx
+++ b/src/components/Table/Internal/Body/Scroller.tsx
@@ -25,6 +25,7 @@ import styles from '../octable.module.scss';
 const BUTTON_HEIGHT: number = 36;
 const BUTTON_PADDING: number = 2;
 const DEFAULT_SCROLL_WIDTH: number = 100;
+const SCROLL_TOLERANCE: number = 1;
 
 export const Scroller = React.forwardRef(
   <RecordType,>(
@@ -138,14 +139,14 @@ export const Scroller = React.forwardRef(
             .reverse()
             .find(
               (leftOffset: number) =>
-                leftOffset > scrollBodyRef.current.scrollLeft
+                leftOffset > scrollBodyRef.current.scrollLeft + SCROLL_TOLERANCE
             );
         } else {
           scrollLeft = scrollOffsets
             .map((offset) => -offset)
             .find(
               (leftOffset: number) =>
-                leftOffset < scrollBodyRef.current.scrollLeft
+                leftOffset < scrollBodyRef.current.scrollLeft - SCROLL_TOLERANCE
             );
         }
       } else {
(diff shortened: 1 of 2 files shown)


src/components/Table/Internal/Body/Scroller.tsx
```diff
@@ -25,6 +25,7 @@ import styles from '../octable.module.scss';
 const BUTTON_HEIGHT: number = 36;
 const BUTTON_PADDING: number = 2;
 const DEFAULT_SCROLL_WIDTH: number = 100;
+const SCROLL_TOLERANCE: number = 1;
```
agupta-eightfold, line 28: Can you apply same tolerance to fix  end arrow stays visible/clickable with nothing left?

Apply same changes there as well to fix it.

agupta-eightfold (approved):
Looks good overall, just a minor fix pending, please check.

src/components/Table/Internal/Tests/Scroller.test.tsx
```diff
@@ -0,0 +1,126 @@
+import React from 'react';
+import Enzyme, { mount, ReactWrapper } from 'enzyme';
+import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
+import MatchMediaMock from 'jest-matchmedia-mock';
+import { Scroller } from '../Body/Scroller';
+import { Button } from '../../../Button';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+let matchMedia: any;
```
agupta-eightfold: Let's not use any and define proper types.

src/components/Table/Internal/Tests/Scroller.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+        stickyOffsets={stickyOffsets}
+        direction={direction}
+        titleRef={titleRef}
+        scrollLeftAriaLabelText="Scroll left"
+        scrollRightAriaLabelText="Scroll right"
+      />
+    );
+  };
+
+  const clickArrow = (wrapper: ReactWrapper, ariaLabel: string): void => {
```
agupta-eightfold, line 88: Simply use simulate('click')?

PR #392, opened 2022-09-27
fix: update snaps to fix broken build

ychhabra-eightfold:
## SUMMARY:
fix: update snaps to fix broken build

## CHANGE TYPE:

-   [ ] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

-   [ ] Tests for this change already exist
-   [ ] I have added unittests for this change

PR #391, opened 2022-09-27
fix: table: add page sizes prop to table pagination

dkilgore-eightfold:
## SUMMARY:
The Table pagination props didn't yet support pageSizes, this change adds support.

https://user-images.githubusercontent.com/99700808/192438187-b700821b-64ce-49d7-803c-0bf90ecee3ca.mp4

## GITHUB ISSUE (Open Source Contributors)
https://github.com/EightfoldAI/octuple/issues/388

## JIRA TASK (Eightfold Employees Only):
ENG-30331

## CHANGE TYPE:

-   [X] Bugfix Pull Request
-   [ ] Feature Pull Request

## TEST COVERAGE:

-   [X] Tests for this change already exist
-   [X] I have added unittests/snapshots for this change

## TEST PLAN:
pull the pr branch and do `yarn` and `yarn storybook` verify the Table Page Sizes story behaves as expected.

PR #1149, opened 2026-07-09
fix(Select/List): improvedA11y combobox VPAT fixes (Hub #242, #391, #392)

agupta-eightfold:
## Summary

Fixes three WAI-ARIA APG combobox a11y defects confirmed live on QA (Allyant Hub #242, #391, #392), all rooted in the shared `Select`/`Menu`/`List` components in `improvedA11y` mode. A single library fix resolves them for every `improvedA11y` `Select` consumer (`TemplateSelectModal`, `TimeSlotSelect`, and any other usage).

### 1. Listbox missing accessible name (#391)
`List` destructured only `aria-label` and applied it to the `<ul>`/`<ol>`; `aria-labelledby` fell into `...rest` and landed on the outer wrapper `<div>` instead of the `role="listbox"` element. Consumers passing `aria-labelledby` via `menuProps` (e.g. `TimeSlotSelect`) got a listbox with no accessible name. Now forwarded to both `<ul>` and `<ol>` alongside `aria-label`.

### 2. `aria-selected` doesn't follow keyboard focus (#392, #242)
`aria-selected` was set once from the committed value (`opt.selected`) and never moved as Arrow keys changed the active descendant, so focus and selection state diverged. Per the APG Combobox pattern ("selection follows focus in the listbox"), **single-select** listboxes now track the active descendant. **Multi-select** listboxes keep `aria-selected` on the committed choices (guarded with `!multiple`) — moving it with focus there would be incorrect.

### 3. VoiceOver never announces the highlighted option (#392, #242)
The visually-hidden live region reacted only to `dropdownVisible`/`options`/`searchQuery`/`selectLocale`, never to `activeDescendantId`, so VoiceOver announced only "expanded" and nothing as the user arrowed through options (NVDA compensates via native `aria-activedescendant`). The active option's text and position are now announced.

**⚠️ Reviewer flag:** this causes NVDA to hear the option name twice (once natively via activedescendant, once via this echo) — accepted tradeoff for VoiceOver support.

The "of" connector in the announcement ("Apple, 1 of 3") is localizable via a new optional `optionPositionSeparatorText` locale key (set on `en_US`/`en_GB`; other locales fall back to `"of"` in code pending translation).

## Test plan
- New unit tests for all three fixes (List: attr lands on list el not wrapper; Select: selection-follows-focus single-select, multi-select unchanged, live-region echo on arrow nav).
- `npx jest src/components/List src/components/Select src/components/Menu` → **122 tests pass**, snapshots clean.
- Verified end-to-end chain: `menuProps.aria-labelledby` → Select `restMenuProps` → Menu `...rest` → List → `<ul role="listbox">`.

## Consumer impact
No app-side code changes needed. `vscode` consumers already pass correct attributes; they only need to bump `@eightfold.ai/octuple` after this releases.

### DIFF
--- a/src/components/Select/Locale/en_GB.tsx
+++ b/src/components/Select/Locale/en_GB.tsx
@@ -6,6 +6,7 @@ const locale: SelectLocale = {
     resultAvailableText: 'match found.',
     resultsAvailableText: 'matches found.',
     noResultsFoundText: 'No match found for',
+    optionPositionSeparatorText: 'of',
   },
 };
 
--- a/src/components/List/List.test.tsx
+++ b/src/components/List/List.test.tsx
@@ -187,4 +187,34 @@ describe('List', () => {
     expect(item2).not.toHaveFocus();
     expect(item1).toHaveFocus();
   });
+
+  test('applies aria-labelledby to the list element (ul), not the wrapper div', () => {
+    const { container } = render(
+      <List
+        {...listProps}
+        listType="ul"
+        role="listbox"
+        aria-labelledby="external-label"
+      />
+    );
+    const list = container.querySelector('ul');
+    const wrapper = container.querySelector('div.my-ref-class');
+    expect(list).toHaveAttribute('aria-labelledby', 'external-label');
+    expect(wrapper).not.toHaveAttribute('aria-labelledby');
+  });
+
+  test('applies aria-labelledby to the ordered list element (ol), not the wrapper div', () => {
+    const { container } = render(
+      <List
+        {...listProps}
+        listType="ol"
+        role="listbox"
+        aria-labelledby="external-label"
+      />
+    );
+    const list = container.querySelector('ol');
+    const wrapper = container.querySelector('div.my-ref-class');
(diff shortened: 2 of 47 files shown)


src/components/Select/Locale/en_GB.tsx
```diff
@@ -6,6 +6,7 @@ const locale: SelectLocale = {
     resultAvailableText: 'match found.',
     resultsAvailableText: 'matches found.',
     noResultsFoundText: 'No match found for',
+    optionPositionSeparatorText: 'of',
```
ychhabra-eightfold, line 9: Need to populate all locals

ychhabra-eightfold (approved):

src/components/Select/Locale/en_GB.tsx
agupta-eightfold: done for all languages.

agupta-eightfold:
Tracked in **ENG-202332** — https://eightfoldai.atlassian.net/browse/ENG-202332 (TA Everest - Sprint 40).

PR #1152, opened 2026-07-13
fix(Select): announce result count with the active option for improvedA11y combobox

shubham-kumar1-eightfold:
## SUMMARY:

The `improvedA11y` combobox echoed the highlighted option into the result-count live region, **overwriting** the count — so screen readers stopped announcing "N matches found" while filtering (regression surfaced after #1149). This combines the count and option in one polite announcement instead.

**What screen readers now hear** (single live region):

- **Open / filter:** `"{count} matches found. {option}, {group}, {n} of {total}"` — one announcement, including the seeded option's **group**.
  - VoiceOver merges live-region updates into one utterance, so racing separate count/option regions dropped one of them.
  - VoiceOver doesn't read the *seeded* option's group natively on open, so we include it here.
- **Arrow navigation:** `"{option}, {n} of {total}"` — option only; **no count** (avoids re-announcing it as noise) and **no group** (VoiceOver reads the option + its group natively while arrowing, so echoing them would double-announce).
- **Debounced** on open/filter so per-keystroke updates coalesce into one announcement and don't race the combobox input's own typing echo; arrow-nav announces immediately.
- **Plain (non-`improvedA11y`) Selects are completely unchanged** — all of the above is gated behind `improvedA11y`.

**Implementation** (minimal delta on #1149):

- `liveRegionMessage` now **appends** the option to the count instead of overwriting it, gated by an `announceModeRef` (`'list'` on open/filter, `'nav'` on arrow — set in the seed effect and `moveActiveDescendant`).
- Group comes from an `optionId → group` map **memoized on `options`** (O(1) lookup, no per-announcement rescan), also gated to `improvedA11y`.
- Debounce is a `useEffect` + `setTimeout` (the repo's standard pattern, e.g. `Accordion`), gated to `improvedA11y`; the delay is the `ANNOUNCE_DEBOUNCE_DURATION` constant in `Select.types.ts`, matching the repo's timing-constant convention (`ANIMATION_DURATION`, …).
- Adds one exported timing constant (`ANNOUNCE_DEBOUNCE_DURATION`), consistent with the existing ones (`ANIMATION_DURATION`, …); no new component props, no locale changes.

N/A

https://eightfoldai.atlassian.net/browse/ENG-203165
Follow-up to the combobox VPAT work in #1149.

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:

- Updated `Select.test.tsx`: open announces the count and `Apple, 1 of 3`; after `ArrowDown` the region announces `Banana, 2 of 3` without the count.
- Added a grouped test: open announces the seeded option's group (`Apple, With HRIS`); `ArrowDown` announces `Banana, 2 of 2` **without** the group (left to VoiceOver's native reading).
- Full `Select.test.tsx` suite passes (81 tests); `Form` suite unaffected (the `improvedA11y` gate keeps plain Selects untouched).
- Verified with VoiceOver + Safari on grouped and ungrouped template Selects.

### DIFF
--- a/src/components/Select/Select.test.tsx
+++ b/src/components/Select/Select.test.tsx
@@ -1514,7 +1514,7 @@ describe('Select improvedA11y', () => {
     });
   });
 
-  test('announces the highlighted option in the live region as focus moves', async () => {
+  test('announces the result count with the highlighted option, and only the option on arrow keys', async () => {
     const { getByPlaceholderText, getByRole, getAllByRole } = render(
       <Select
         improvedA11y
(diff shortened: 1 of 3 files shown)


agupta-eightfold (approved):
lgtm

PR #1158, opened 2026-08-08
fix(react19): remove React 19 incompatible APIs and restore compatibility

kkarthik-eightfold:
## SUMMARY:

Makes `@eightfold.ai/octuple` compatible with React 19 so it can be consumed by Talent Forge, by removing all removed/incompatible React 19 APIs while preserving public APIs and existing behavior.

React 19 removed `ReactDOM.render`, `ReactDOM.unmountComponentAtNode`, and `ReactDOM.findDOMNode`, and moved element refs from a top-level `.ref` property onto `props.ref`. These broke Octuple in several places:

- **`DialogHelper`** used `ReactDOM.render`/`unmountComponentAtNode` directly, which made the whole `@eightfold.ai/octuple` barrel import fail under React 19. Migrated to `createRoot`/`root.render`/`root.unmount` from `react-dom/client`, tracking one root per container id so `close()` unmounts the right instance. Public `show()`/`showSmall()`/`showMedium()`/`close()` APIs are unchanged.

- **`findDOMNode.ts`, `SingleObserver.tsx`, `CSSMotion.tsx`, `Trigger.tsx`, `useHeights.tsx`** all depended on `ReactDOM.findDOMNode` (directly or via the shared `findDOMNode()` helper) to resolve a DOM node from a component that couldn't accept a ref. `findDOMNode()` is now a null-safe passthrough with no `ReactDOM` dependency. `DomWrapper` (the `SingleObserver`/`CSSMotion` fallback) now resolves the DOM node via a hidden marker node instead, placed *after* (not before) its children so `container.firstChild`/`:first-child` semantics are preserved; the marker is only rendered at all when a direct ref genuinely can't be attached (see `canAttachRef` in `ref.ts`), so `Table`'s `<td>` (which must stay a direct child of `<tr>`) and other host/`forwardRef` children get zero extra DOM nodes.

- Added `canAttachRef()` and `getElementRef()` to `ref.ts`. `getElementRef()` reads `props.ref` on React 19 and falls back to the legacy top-level `.ref` on React 16.8–18 (Octuple's `peerDependencies` still span that range) — reading `props.ref` alone silently drops real refs on React <19, which broke `Align`'s `forceAlign` wiring inside `PopupInner` until this was corrected. Rolled out across `SingleObserver`, `CSSMotion`, `Trigger`, `Align`, and `useItemRef`.

- `react-is@18` doesn't recognize React 19 elements (different `$$typeof` symbol), and `react-is@19` doesn't recognize React <19 elements — `react-is` versions are generation-locked, not version-range compatible. Since a single bundled `react-is` version can't correctly serve every React major Octuple supports, `toArray.ts`'s Fragment detection and `ref.ts`'s `forwardRef` detection now check the relevant symbols directly instead of delegating to `react-is`. Bumped `react-is`/`@types/react-is` to 19.x per the target anyway.

- `@react-spring/web`'s `peerDependencies` already allow React 19; left unchanged.

- Added a `global.d.ts` ambient declaration for `react-dom/client` so the build type-checks against this repo's own React 17 devDependency, which predates that module.

N/A

N/A

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

Added: `DialogHelper.test.tsx` (mount/unmount via `createRoot`), `findDOMNode.test.ts` (null-safety), `reactIs.test.tsx` (React 19 Fragment detection via `toArray`), plus rewritten cases in `CSSMotion.test.tsx` (DOM resolution without `findDOMNode`). Updated 35 pre-existing snapshots (Select, SkillBlock, DatePicker) that gained the new hidden `DomWrapper` marker node, after confirming each is the accepted, documented trade-off.

## TEST PLAN:

- `yarn typecheck` — pass
- `yarn lint` — pass
- `yarn test` — 225/225 suites, 2657/2679 tests pass (22 pre-existing skips), 441/441 snapshots pass
- `yarn build` (rollup) — succeeds, including the `react-dom/client` import
- Manually verified no remaining `ReactDOM.render`/`unmountComponentAtNode`/`ReactDOM.findDOMNode`/legacy `element.ref` usage in `src`

### Known limitations / follow-ups

- This repo's own devDependencies remain on React 17 (bumping them breaks the 144-file Enzyme-based test suite, which has no official React 18/19 adapter), so `DialogHelper.test.tsx` uses a virtual `jest.mock('react-dom/client', ...)` shim backed by React 17's `ReactDOM.render`/`unmountComponentAtNode` to exercise the public API against this repo's own toolchain. Real consumers on React 18/19 get the real module. A follow-up to migrate the test suite off Enzyme would let this repo verify against a real React 19 install directly.
- Found and flagged separately (not fixed here, out of scope): `Dialog.tsx` accepts a `parent` prop but never forwards it to `BaseDialog`, so `DialogHelper` always portals into `document.body` regardless of the container it creates. Pre-existing, unrelated to React 19.
- 6 components (`Tree`, `DirectoryTree`, `BaseTree`, `Table`, `OcTable`, plus `TreeNode`'s class component) set `.defaultProps` on function/forwardRef components. React 19 still supports this but logs a deprecation warning; not addressed here since it doesn't break functionality and wasn't part of the original investigation's scope.


agupta-eightfold:
Closing in favor of #1161, which supersedes this. Thanks for the groundwork here, several pieces of this PR live on there with credit: the dual-symbol Fragment detection in `toArray`, the `props.ref`-first read order in `getElementRef`, and parts of the test scaffolding.

Why we changed approach. The hard requirement for this fix was zero behavior change for React 16/17/18 consumers (vscode `www/react` is React 16.14 + webpack 5 with octuple on a caret range), and a few of the choices here broke that:

1. The static `import { createRoot } from 'react-dom/client'` in DialogHelper ships in the bundle. `react-dom/client` does not exist in react-dom 16/17, so webpack consumers fail with module-not-found on their next octuple bump. The Storybook shim this PR adds is the tell: every React <18 webpack consumer would need the same shim, and even shimmed, React 16/17 has no `createRoot` at runtime. #1161 uses property access plus an `unstableSetRender` registration API instead, so octuple never references `react-dom/client`.
2. The hidden marker `<span>` in `DomWrapper` renders on every React version, which changes rendered DOM on React 17 (the 35 updated snapshots in this diff are that change). CSS structural selectors and e2e selectors in consumers are exposed to it. #1161 gates the marker so it only renders when `ReactDOM.findDOMNode` is actually absent, and its full suite passes with zero snapshot updates.
3. `Trigger.getRootDomNode` returning null where `ReactDOM.findDOMNode(this)` used to resolve regresses Tooltip/Dropdown positioning for class components and plain function children on old React too. Similarly, `SingleObserver` can now hand a class instance straight to `ResizeObserver.observe`, which throws. #1161 keeps the guarded legacy fallbacks on <=18.
4. The `defaultProps` note in the description is not quite right: React 19 removed (not deprecated) `defaultProps` for function components, so Table loses `rowKey: 'key'` and `emptyText`, and Tree flips `showIcon`. #1161 converts those six components to destructure defaults.

None of this diminishes the investigation here, it mapped the right problem surface and #1161 built on it.

PR #1168, opened 2026-09-21
fix(Trigger): popup animation restarts while it is measured and aligned (ENG-212057)

ayush-raj-eightfold:
## SUMMARY:

The popup's entry and exit animations restart mid-flight, so the panel visibly flickers as it opens and as it closes. The only consumer of `Trigger` in the library is `DateTimePicker`, so this shows up on `DatePicker` / `RangePicker`.

`OcPickerTrigger` puts the `slide-up-enter` / `slide-up-leave` classes — which carry the `slideUpIn` / `slideUpOut` **keyframes** — on the popup purely from `visible`. Separately, `PopupInner` + `useVisibleStatus` run a `measure → alignPre → align → motion` pipeline whose `setStatus` calls happen inside `requestAnimationFrame`.

Under React <= 17 those rAF updates were flushed synchronously, so the pipeline finished before the first paint and the keyframes ran exactly once, already positioned. Under React 18 they are concurrent updates rendered in later scheduler tasks, so the popup is painted — and its keyframes start — while it is still being measured and aligned. Each measure/align pass hides and moves the node (`display: none`, `left: -9999px`), and **hiding an element restarts its CSS keyframe animation**. The same happens on the way out when a host re-render triggers a re-align during the exit animation.

A third, related artifact: when the exit animation ends, `useStatus` clears the motion status and style. Under React 18 that is committed and painted one frame before the owner actually hides the node, so a popup that has just finished fading out is shown again at full size and opacity for a frame.

### Changes

1. `PopupInner` — `animationName: 'none'` alongside the existing `opacity: 0`, under the same condition, so the keyframes cannot start before the node is measured and aligned. `animationName` rather than the `animation` shorthand, which would also reset duration, timing and fill mode.
2. `PopupInner` — `alignDisabled` also requires `visible`, so a popup animating out is never re-aligned.
3. `PopupInner` — the `alignTimes` layout effect gates on `visible` too. Aligning is what advances `alignTimes`, so without this, (2) stalls the status pipeline at `align` during leave and `pointer-events: none` is never cleared.
4. `useStatus` — `flushSync` the motion-end `setStatus(STATUS_NONE)` + `setStyle(null)`.

No public API change.

NA

https://eightfoldai.atlassian.net/browse/ENG-212057

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:

1. Switch to hm2@eightfolddemo-gsawlani.com
2. Go to /feedback/dashboard?view=interviewer&status=REQUESTED
3. Check date range filer, use quick date range filter (Last 90, 60, 30 days..) 
4. Regression - Ensure the popup opens and you're able to select date range using the filters
5. Ensure there is no flickering when the date range picker opens


agupta-eightfold (approved):

src/components/Trigger/Popup/PopupInner.tsx
```diff
@@ -144,14 +144,18 @@ const PopupInner = forwardRef<PopupInnerRef, PopupInnerProps>((props, ref) => {
     zIndex,
     opacity:
       status === 'motion' || status === 'stable' || !visible ? undefined : 0,
+    animationName:
```
agupta-eightfold, line 147: `animationName: none` only applies before status reaches `motion`. At `stable` it is undefined and `OcPickerTrigger.tsx:76` still has `slideUpEnter` on the node, so any re-align while open (preset click, window resize) replays the slide-in via `dom-align`'s `display:none` restore. Can we tie the enter/leave classes to the motion phase instead of `visible`?

src/components/Motion/hooks/useStatus.ts
```diff
@@ -82,8 +83,10 @@ export const useStatus = (
 
     // Only update status when `canEnd` and not destroyed
     if (status !== STATUS_NONE && currentActive && canEnd !== false) {
-      setStatus(STATUS_NONE, true);
-      setStyle(null, true);
+      flushSync(() => {
```
agupta-eightfold: This hook drives every `CSSMotion`, not just the popup, and `onLeaveEnd` runs before the `flushSync` so its pending update is flushed too. A `CSSMotionList`/`UploadList` with N items leaving becomes N blocking renders instead of one batch. Could this live in `PopupInner` where `visible` is known?

src/components/Trigger/Tests/motion.test.jsx
```diff
@@ -105,4 +105,24 @@ describe('Trigger.Motion', () => {
       pointerEvents: 'none',
     });
   });
+
+  it('popup should not animate until it is measured and aligned', () => {
```
agupta-eightfold: This asserts the same initial-render state the updated snapshots already cover, and nothing is asserted after `jest.runAllTimers()` on L126. If the pipeline ever stalls before `motion`, `animation-name: none` sticks and the picker silently stops animating. Worth asserting it is released at `stable`, plus a case for the `alignDisabled` change.

src/components/Trigger/Popup/PopupInner.tsx
ayush-raj-eightfold: You're right, and it reproduces. I measured it with a per-animation-frame opacity recorder on a page with a `RangePicker`: open the popup, let it settle, then force a re-align by resizing the viewport. Frames below full opacity while the popup is open:

| build | frames |
| --- | --- |
| React 18 + octuple 2.59.0 | 20 |
| React 18 + this branch | 20 |

So it is real, my change does nothing for it, and it is equally present on `main` — it is not part of the React 18 regression this PR is about. Root cause is the one you named: `dom-align`'s `forceRelayout()` sets `display: none` to force a reflow, and the node permanently carries `slideUpEnter`, so every align replays the keyframes.

I tried your suggestion and backed it out. Tying the classes to the motion phase means handing the motion to `CSSMotion` — `popupMotion={{ motionName: { enter, enterActive, leave, leaveActive } }}` (the map form, since these are CSS-module class names) plus `-enter-active` / `-leave-active` rules with `animation-play-state: paused | running`. That works, but it changes the close lifecycle: with a real `motionName`, `CSSMotion` keeps the popup mounted until `animationend`. **17 picker tests fail** on `isClosed()`, and in production the popup would not close at all if `animationend` never fires (reduced motion, a consumer's `animation: none`, a background tab) unless a `motionDeadline` is added too. It also needs a guard in `PopupInner`, which merges `motion?.motionName` straight into `mergeClasses` — an object there injects literal `enter`/`enterActive` classes.

What I could not do is fix it narrowly. Re-arming the gate at `stable` kills the animation: with no `motionName` today, `PopupInner`'s `if (!motion.motionName && status === 'motion') goNextStatus()` reaches `stable` one rAF after `motion`, long before the 200ms keyframes finish.

So this wants to be its own PR — motion-phase classes, `motionDeadline`, the `mergeClasses` guard, and the picker test rework. Happy to raise it; shall I open an issue, or do you want it folded in here?

src/components/Motion/hooks/useStatus.ts
ayush-raj-eightfold: Fair — fixed in 8b9c04f. The flush is now opt-in via a `flushOnMotionEnd` prop on `CSSMotionProps`, and `PopupInner` is the only thing that sets it. `CSSMotionList`, `UploadList` and every other consumer are untouched, so N items leaving still batch into one render.

It has to stay in `useStatus` rather than move into `PopupInner`: the frame comes from `setStatus(STATUS_NONE)` + `setStyle(null)` themselves, and `PopupInner`'s `onLeaveEnd` wrapper runs *before* those, so it cannot flush them.

On why the flush is needed at all — `animationend` is not a discrete event, so React 18 gives it `DefaultLane` and renders it in a later scheduler task. The animation has already ended, so the element is back to its un-animated state (no `fill-mode: forwards`), but the owner has not applied `leavedClassName` yet: one painted frame at full size and opacity. Measured, 4 open/close cycles on React 18: 4 of those frames without the flush, 0 with it.

I left `flushOnMotionEnd` out of `MOTION_PROP_NAMES` deliberately, so `CSSMotionList` cannot pass it down to its children.

src/components/Trigger/Tests/motion.test.jsx
ayush-raj-eightfold: Rewritten in 8b9c04f. It now asserts the gate is applied **and released** together with the pre-existing `opacity` gate, across a `popupVisible` flip — which the snapshots do not cover, since they only capture the open state:

```jsx
expect(popup().style.opacity).toBe('0');
expect(popup().style.animationName).toBe('none');
rerender(genTrigger({ popupVisible: false }));
expect(popup().style.opacity).toBe('');
expect(popup().style.animationName).toBe('');
```

Verified it discriminates: fails against `upstream/main`, passes here.

On "it sticks if the pipeline stalls" — it can only stick where `opacity: 0` already sticks. Both are gated on the identical `status === 'motion' || status === 'stable' || !visible` expression, so there is no state where the popup is visible but frozen. A stalled pipeline was already fatal before this PR; it made the popup invisible, and now it is invisible and unanimated.

Two things I could not test here, rather than fake them:

1. **Released at `stable`.** The pipeline never leaves the pre-`motion` phase in jsdom — `alignTimes` only advances when `dom-align` actually aligns, which needs layout. That is why the existing `Trigger.Util > mock > open` expects `opacity: 0` permanently. I tried real timers and `waitFor` (the `FadeIn` pattern); it still stalls. The coupling assertion above is the testable form of the same invariant.
2. **The `alignDisabled` case.** Same reason: `status` never reaches `align`, so `disabled` is already `true` throughout leave on `main`. I wrote the test, saw it pass with *and* without the fix, and deleted it rather than ship a green-either-way test. That change is covered by the browser measurements in the PR body (4 fade-out restarts → 0).

If you'd rather have jsdom coverage for those two, it needs `getBoundingClientRect`/`offsetParent` stubs so `dom-align` can complete — happy to add that as a test-infra change if you want it.

PR #1169, opened 2026-09-22
fix(Tooltip, FocusTrap, Select): fix show/hide race, focus stealing on trap activation, and phantom onOptionsChange call

ayadav-eightfold:
## SUMMARY:

Three independent bug fixes found while investigating a component stuck in an incorrect visual state under fast, back-to-back interactions:

**1. `Tooltip`: fixed a race that could leave a hover/focus-triggered tooltip permanently visible**

`toggle()` schedules both showing and hiding through a single delayed timer (to allow the fade animation to play before content is removed). If a blur/mouseleave arrives *before* a pending show commits, the old `onBlur`/`onMouseLeave` handlers were gated on the not-yet-updated `mergedVisible` state, so the close handler never even attached — the blur event was silently dropped, and the pending show committed a moment later with nothing left to reverse it, leaving the tooltip stuck open indefinitely.

Fix:
- Added `intendedVisibleRef`, tracking the requested visible state synchronously the instant `toggle()` runs, instead of relying on the delayed `mergedVisible`.
- Converted the internal `timeout` from a plain per-render local variable to a `useRef`, so a close handler firing on a later render can actually cancel a timer scheduled by an earlier render's closure (a per-render variable can never reach back and do that).
- Restructured `onBlur` into an always-attached handler that reads the ref at invocation time, rather than a render-time ternary — necessary because a same-value `setHiding` call (e.g. the very first show) causes React to skip re-rendering, which would otherwise leave a stale handler attached regardless of the ref's value.

**2. `FocusTrap`: fixed `useFocusTrap` stealing focus from an already-correctly-focused element on activation**

On activation, `setUpFocus` unconditionally force-focused `firstFocusableSelector` (or the first focusable element), regardless of where focus currently was. This breaks the common `onFocus={() => setTrapActive(true)}` activation pattern: the trap turns on *because* something inside it was just explicitly focused, but the activation effect then ignored that and redirected focus elsewhere.

Fix: skip the force-focus step when `document.activeElement` is already inside the trap's container. The existing "activate with focus outside the container" behavior (the common case) is unaffected.

**3. `Select`: fixed `onOptionsChange` firing once with an empty selection before firing again correctly**

The internal `options` state was initialized with every option's `selected` hardcoded to `false`, regardless of `defaultValue` — the correct selection was only applied later via a `[defaultValue]`-keyed effect. Since the effect that calls `onOptionsChange` runs on every render where the selection changes (including the first), consumers saw one call with an empty selection immediately on mount, followed by a second, corrective call once the `defaultValue` effect caught up.

Fix: compute `selected` from `defaultValue` directly in the `options` initializer (mirroring the existing `[defaultValue]` effect's own logic), and likewise compute the initial displayed text from that already-correct `options` state. `onOptionsChange` now fires once, correctly, from the first render.

https://eightfoldai.atlassian.net/browse/ENG-207605

## CHANGE TYPE:

- [x] Bugfix Pull Request

## TEST COVERAGE:

## TEST PLAN:

- Added regression tests for all three fixes:
  - `Tooltip.test.tsx`: blur arriving before a pending show commits no longer leaves the tooltip stuck visible.
  - `useFocusTrap.test.tsx`: activating the trap via an explicit focus already inside it no longer redirects focus to the first focusable element.
  - `Select.test.tsx`: `onOptionsChange` fires exactly once on mount with the correct `defaultValue` selection, not `[], []` first. Also removed an existing assertion that had encoded the old (buggy) two-call sequence as expected behavior.
- Full existing suite verified green after each change: 230 test suites / 2703 tests / 441 snapshots, 0 failures.
- `tsc --noEmit` clean.


src/components/Tooltip/Tooltip.tsx
```diff
@@ -655,7 +663,11 @@ export const Tooltip: FC<TooltipProps> = React.memo(
               event: React.MouseEvent<HTMLDivElement, MouseEvent>
```
agupta-eightfold, line 680: `mergedVisible` also tracks the `visible` prop, but `intendedVisibleRef` is only written inside `toggle`. So a tooltip opened via the `visible` prop never gets mouseleave/blur, and `onVisibleChange(false)` never fires. Verified against main: `[[false]]` there, `[]` here. Can we gate on `(intendedVisibleRef.current || mergedVisible)`?

src/components/Select/Select.tsx
```diff
@@ -152,7 +152,11 @@ export const Select: FC<SelectProps> = React.forwardRef(
 
     const [options, setOptions] = useState<SelectOption[]>(
       (_options || []).map((option: SelectOption, index: number) => ({
-        selected: false,
+        selected:
```
agupta-eightfold: `selected` is computed before `...option` spreads, so options that already carry `selected` overwrite it and still get the double `onOptionsChange`. The effect at line 283 does it right with `|| opt.selected`. Compute after the spread and OR it, and add a fixture with `selected` since the current one has none.

src/shared/FocusTrap/hooks/useFocusTrap.ts
```diff
@@ -68,6 +68,13 @@ export function useFocusTrap(
       return;
     }
     restoreFocusRef.current = document.activeElement;
+
+    // If focus is already somewhere inside the trap -- respect it instead of redirecting
```
agupta-eightfold: The early return skips the `firstFocusableSelector` lookup and its retry interval, so with an `autoFocus` child the selector is ignored (`auto` gets focus here, `preferred` on main). Consumers passing this prop lose it silently. Return early only when no selector was given, or when activeElement is already the resolved element.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -655,7 +663,11 @@ export const Tooltip: FC<TooltipProps> = React.memo(
...
             ): void => {
               const floatingElement: HTMLElement = refs.floating.current;
-              if (trigger.includes('hover') && mergedVisible && !gestureType) {
+              if (
+                trigger.includes('hover') &&
+                intendedVisibleRef.current &&
+                !gestureType
+              ) {
                 toggle(
                   !event.currentTarget &&
```
agupta-eightfold, line 689: `onMouseEnter`/`onFocus` are still gated on `!mergedVisible`, so re-entering the trigger inside the `hideAfter` window finds no handler, the pending hide is not cancelled and the tooltip vanishes under the pointer. Same shape as the fix above, worth doing both directions in one go.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -126,7 +131,9 @@ export const Tooltip: FC<TooltipProps> = React.memo(
         `${tooltipId?.current}-wrapper`
       );
```
agupta-eightfold, line 142: The comment says timeouts are preserved for unmount, but there is no unmount cleanup, so the pending timer still calls `setVisible`/`onVisibleChange` after the component is gone. One line fixes it: `useEffect(() => () => clearTimeout(timeoutRef.current), []);`

src/shared/FocusTrap/hooks/useFocusTrap.ts
```diff
@@ -68,6 +68,13 @@ export function useFocusTrap(
       return;
     }
     restoreFocusRef.current = document.activeElement;
```
agupta-eightfold: `restoreFocusRef.current` is assigned before the new early return, so on the focus-already-inside path the restore target sits inside the trap and focus drops to `<body>` on close. Assign it only when activeElement is outside the container.

src/components/Select/Select.tsx
ayadav-eightfold: Fixed in ce42cad1 -- `selected` is now computed after the `...option` spread (mirroring the `[defaultValue]` effect's own pattern) and coerced with `!!(...)` so it never leaks `undefined`. Added a regression test with a fixture that carries a raw `selected` field.

src/shared/FocusTrap/hooks/useFocusTrap.ts
ayadav-eightfold: Fixed in ce42cad1 -- the early return now only skips the redirect when there's no `firstFocusableSelector`, or when `document.activeElement` is already the resolved target. Added a regression test that mirrors a native `autoFocus` child grabbing focus first.

src/shared/FocusTrap/hooks/useFocusTrap.ts
ayadav-eightfold: Fixed in ce42cad1 -- `restoreFocusRef.current` is now only assigned when `document.activeElement` is outside the trap's container, so activating via focus already inside no longer captures a bogus restore target. Added a regression test.

src/components/Tooltip/Tooltip.tsx
ayadav-eightfold: Fixed in ce42cad1 -- `intendedVisibleRef` is now kept in sync with the controlled `visible` prop via a `useEffect`, so a tooltip opened purely via `visible` (no `toggle()` call) still fires `onVisibleChange(false)` on blur/mouseleave. Added a regression test.

src/components/Tooltip/Tooltip.tsx
ayadav-eightfold: Fixed in ce42cad1 -- `onMouseEnter`/`onFocus` are now gated on `intendedVisibleRef.current` (matching `onMouseLeave`/`onBlur`) instead of the async-stale `mergedVisible`, so re-entering within the `hideAfter` window correctly cancels the pending hide. Added a regression test that re-enters mid-window and asserts the tooltip stays visible.

src/components/Tooltip/Tooltip.tsx
ayadav-eightfold: Fixed in ce42cad1 -- added `useEffect(() => () => clearTimeout(timeoutRef.current), [])` so a pending timer can no longer call `setVisible`/`onVisibleChange` after unmount.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -647,15 +666,21 @@ export const Tooltip: FC<TooltipProps> = React.memo(
             }}
             onKeyDown={!gestureType ? handleReferenceKeyDown : null}
             onMouseEnter={
-              trigger.includes('hover') && !mergedVisible && !gestureType
+              trigger.includes('hover') &&
+              !intendedVisibleRef.current &&
```
ayadav-eightfold: `onMouseEnter` (L668-672) and `onFocus` (L691-695) read `intendedVisibleRef.current` at render time, but a ref mutation never triggers a re-render -- this is exactly why you restructured `onBlur` into an always-attached handler at L698. Concretely: the sync effect at L123-127 writes the ref *after* the render commits, so on a controlled `visible={true}` mount these two are computed with the ref still `false` and stay attached; hovering or focusing the already-open trigger then runs `toggle(true)` and fires a duplicate `onVisibleChange(true)` that the old `!mergedVisible` guard suppressed. Make both always-attached handlers that read the ref at invocation, like `onBlur`/`onMouseLeave`.

src/shared/FocusTrap/hooks/useFocusTrap.ts
```diff
@@ -67,11 +67,32 @@ export function useFocusTrap(
...
-    restoreFocusRef.current = document.activeElement;
+
+    const focusIsAlreadyInside: boolean = elRef.current.contains(
+      document.activeElement
+    );
+
+    // Only capture a restore target when focus is truly outside the trap -- if it's
+    // already inside, there's no prior external focus state worth restoring on close.
+    if (!focusIsAlreadyInside) {
+      restoreFocusRef.current = document.activeElement;
```
ayadav-eightfold, line 80: Skipping the assignment leaves the previous activation's value in `restoreFocusRef.current`, and the cleanup at L111 focuses it unconditionally. Activate from outside (captures A) -> deactivate -> re-activate via focus already inside -> deactivate, and focus jumps back to A even though it was never outside this time. Add `else { restoreFocusRef.current = null; }` -- the new test at useFocusTrap.test.tsx only covers a trap that was never activated before, so it misses this.

ayadav-eightfold (commented):
Three residual issues from the ref/focus rework; details inline.

src/components/Select/Select.tsx
```diff
@@ -152,17 +152,31 @@ export const Select: FC<SelectProps> = React.forwardRef(
 
     const [options, setOptions] = useState<SelectOption[]>(
       (_options || []).map((option: SelectOption, index: number) => ({
-        selected: false,
         hideOption: false,
         id: `${selectMenuId.current}-option-${index}`,
         object: option.object,
         role: 'option',
         'aria-selected': option.selected,
```
ayadav-eightfold: `aria-selected` is still taken from the raw prop while `selected` below is now derived from `defaultValue`, so with `defaultValue='option2'` and no raw `selected` field the option object carries `selected: true` alongside `'aria-selected': undefined`. That object is what `onOptionsChange` hands consumers as its second argument -- compute `aria-selected` from the same expression.

src/components/Select/Select.tsx
```diff
@@ -219,15 +236,20 @@ export const Select: FC<SelectProps> = React.forwardRef(
...
-          selected: !!selected.find((opt) => opt.value === option.value),
-          hideOption: false,
-          id: `${selectMenuId.current}-option-${index}`,
-          object: option.object,
-          role: 'option',
-          'aria-selected': option.selected,
-          ...option,
-        }))
+        (_options || []).map((option: SelectOption, index: number) => {
+          const isSelected = !!selected.find(
```
agupta-eightfold: These two effects still put `selected` and `'aria-selected'` before the `...option` spread, so for options arriving with `selected: true` the spread fixes `selected` but leaves `'aria-selected': false` in the `onOptionsChange` payload (`{selected:true, aria:true}` on main, `aria:false` here). Rendering is unaffected since line 862 recomputes, but worth mirroring the initializer here. The `[defaultValue]` effect at 310 has the same gap.

src/shared/FocusTrap/hooks/useFocusTrap.test.tsx
```diff
@@ -65,4 +65,188 @@ describe('useFocusTrap', () => {
     await waitFor(() => expect(secondButton.matches(':focus')).toBe(true));
     expect(secondButton.matches(':focus')).toBe(true);
   });
+
+  test('setUpFocus does not steal focus when the trap activates via focus already inside it', async () => {
```
agupta-eightfold, line 69: This one passes on main too, where the bug is present: the steal happens on the 100ms `FOCUS_DELAY_INTERVAL` and `waitFor` resolves on its first sync poll, so the assertion runs before the first tick. With a 400ms wait it goes `b1` on main and `b2` here. Can we add the wait so a revert actually turns this red?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -126,7 +140,11 @@ export const Tooltip: FC<TooltipProps> = React.memo(
         `${tooltipId?.current}-wrapper`
       );
 
-      let timeout: ReturnType<typeof setTimeout>;
+      // Preserves timeouts for when the component unmounts or the user toggles visibility quickly, we need to keep a ref to the timeout ID.
+      const timeoutRef: React.MutableRefObject<ReturnType<typeof setTimeout>> =
+        useRef<ReturnType<typeof setTimeout>>(null);
+
+      useEffect(() => () => clearTimeout(timeoutRef.current), []);
```
agupta-eightfold, line 147: Nothing unmounts with a pending timer, so this cleanup can be deleted and the suite stays green. Hovering then unmounting gives `[[true]]` on main and `[]` here, so it is ~10 lines to lock in.

agupta-eightfold (approved):

PR #1171, opened 2026-09-22
fix(MatchScore): clamp score to avoid invalid array length runtime error

kheinrich-eightfold:
## SUMMARY:

Previously, the `MatchScore` component computed its filled/half/empty circle counts directly from an unbounded `score` prop, so a `score` outside the `[0, total]` range (observed in production as `score = -1`) produced a negative circle count and threw `Uncaught RangeError: Invalid array length` on `Array(n)`.

This PR clamps `score` to `[0, total]` before it's used in any of the circle/label calculations, so out-of-range scores degrade gracefully (rendering `0` or `total` circles) instead of crashing:
- Clamp negative `score` values to `0`.
- Clamp `score` values above `total` down to `total`.
- Rename the misleadingly-named `absScore` (it rounded, but never took an absolute value) to `displayScore` for clarity.
- Add regression tests covering negative and above-total scores.

https://eightfoldai.atlassian.net/browse/ENG-215683

## CHANGE TYPE:

- [x] Bugfix Pull Request
- [ ] Feature Pull Request

## TEST COVERAGE:

## TEST PLAN:
- `yarn jest` — full suite passing
- New tests pass
- No behavior change when `score` is a valid value between 0 and `total` (5 by default)


src/components/MatchScore/MatchScore.tsx
```diff
@@ -40,12 +40,15 @@ export const MatchScore: FC<MatchScoreProps> = React.forwardRef(
       : contextualTheme || theme;
 
     const absTotal: number = Math.abs(total);
-    const absScore: number = Math.round(score);
-    const fullCircles: number = Math.trunc(Math.round(score * 2.0) / 2.0);
+    const clampedScore: number = Math.max(0, Math.min(score, absTotal));
```
agupta-eightfold, line 43: `Math.min(NaN, absTotal)` is `NaN`, so `score={NaN}` still reaches `Array(NaN)` and throws the same `RangeError` this PR is fixing. Coercing a non-finite score to 0 before the clamp should cover it.

src/components/MatchScore/MatchScore.tsx
```diff
@@ -40,12 +40,15 @@ export const MatchScore: FC<MatchScoreProps> = React.forwardRef(
...
+    const displayScore: number = Math.round(clampedScore);
+    const fullCircles: number = Math.trunc(
+      Math.round(clampedScore * 2.0) / 2.0
+    );
     const halfCircle: number = Math.trunc(
-      Math.ceil(score - fullCircles - 0.25)
+      Math.ceil(clampedScore - fullCircles - 0.25)
     );
-    const emptyCircles: number = total - fullCircles - halfCircle;
+    const emptyCircles: number = absTotal - fullCircles - halfCircle;
```
agupta-eightfold, line 51: `total` is a plain `number`, hence `total={5.5}` gives `emptyCircles` 2.5 and `Array(2.5)` throws as well. Can we `Math.trunc` the `absTotal`, so the bound matches the circles we actually render?

src/components/MatchScore/MatchScore.test.tsx
```diff
@@ -43,4 +43,16 @@ describe('MatchScore', () => {
     );
     expect(container.querySelector('.label').textContent).toBe('Some label ');
   });
+
+  test('MatchScore clamps a negative score instead of throwing', () => {
```
agupta-eightfold, line 47: Both tests assert only the label, which is computed from `displayScore` and not from `fullCircles`/`halfCircle`/`emptyCircles`. A clamp applied to the label alone would keep them green, so it would help to assert the circle counts too.

src/components/MatchScore/MatchScore.tsx
kheinrich-eightfold: Updated

agupta-eightfold (approved):

